PR #8, opened 2024-11-15
Provide CI testing support

mitchnegus:
This provides a starting point for converting our GitLab CI testing infrastructure to the GitHub Actions workflow. 

### Workflow Upgrades

To avoid duplication in the workflow scripts, I've created a new set of [composite actions](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action) to install minimega, discovery, FIREWHEEL, and tox. We should hopefully be able to use these composite actions to dramatically simplify some of the CI scripts we had before. I've also upgraded the versions of referenced reusable versions where possible.

### Test Enhancements

The unit tests have been upgraded to provide increased utility and work better through the FIREWHEEL interface:
- Running the `firewheel test unit` helper now returns the appropriate exit code returned by pytest, which was previously unintentionally suppressed.
- A new marker includes or excludes tests which depend on model components.
- In service of that functionality, the tests now also make use of [pytest's built-in `-m` option](https://docs.pytest.org/en/stable/example/markers.html#marking-test-functions-and-selecting-them-for-a-run) to select marked tests for inclusion/exclusion rather than adding custom options.

### Bug Fixes

This PR also fixes a few bugs in the current implementation:
- This fixes a testing error identified by testing the FIREWHEEL deployment in an fresh installation (a hardcoded path in the `test_cli_completion.py` script). 
- Arguments passed to the helpers were not parsed correctly, ignoring quoted inputs and just splitting on spaces. This changes the helpers to use `shlex` for proper parsing.

### DIFF
--- a/.github/actions/prepare-discovery/action.yml
+++ b/.github/actions/prepare-discovery/action.yml
@@ -1,0 +1,15 @@
+# A composite action to install discovery
+
+name: Prepare discovery
+
+description: Install discovery
+
+runs:
+  using: composite
+  steps:
+    - name: Install discovery
+      run: |
+        wget https://github.com/mitchnegus/minimega-discovery/releases/download/firewheel-debian_faed761/discovery.deb
+        sudo dpkg -i discovery.deb
+        sudo chown -R $USER:minimega /opt/discovery
+      shell: bash
--- a/.github/actions/prepare-minimega/action.yml
+++ b/.github/actions/prepare-minimega/action.yml
@@ -1,0 +1,29 @@
+# A composite action to install and initialize minimega
+
+name: Prepare minimega
+
+description: Install and initialize minimega
+
+runs:
+  using: composite
+  steps:
+    - name: Install minimega
+      run: |
+        wget https://github.com/sandia-minimega/minimega/releases/download/2.9/minimega-2.9.deb
+        sudo dpkg -i minimega-2.9.deb
+        sudo chown -R $USER:minimega $MM_INSTALL_DIR
+        sudo ln -s $MM_INSTALL_DIR/bin/minimega /usr/local/bin/minimega
+        sudo ln -s $MM_INSTALL_DIR/bin/minimega /usr/local/bin/mm
+      shell: bash
+    - name: Initialize minimega
+      run: |
+        echo -n "" | sudo $MM_INSTALL_DIR/misc/daemon/minimega.init install
+        sudo mkdir -p $(dirname $MINIMEGA_CONFIG)
+        sudo sed -i "s|MINIMEGA_DIR=\"/opt/minimega/\"|MINIMEGA_DIR=\"$MM_INSTALL_DIR/\"|g" $MINIMEGA_CONFIG
+        sudo sed -i "s|MM_RUN_PATH=\"/tmp/minimega\"|MM_RUN_PATH=\"$MM_BASE/\"|g" $MINIMEGA_CONFIG
+        sudo sed -i "s|MM_MESH_DEGREE=0|MM_MESH_DEGREE=1|g" $MINIMEGA_CONFIG
+        sudo sed -i "s|MM_LOG_LEVEL=\"error\"|MM_LOG_LEVEL=\"debug\"|g" $MINIMEGA_CONFIG
+        sudo sed -i "s|MM_LOG_FILE=\"/tmp/minimega.log\"|MM_LOG_FILE=\"$LOG_DIR/minimega.log\"|g" $MINIMEGA_CONFIG
+        sudo systemctl restart minimega
+        sudo chown -R $USER:minimega $MM_BASE $LOG_DIR
+      shell: bash
--- a/.github/actions/prepare-tox/action.yml
+++ b/.github/actions/prepare-tox/action.yml
@@ -1,0 +1,15 @@
+# A composite action to prepare tox-based workflows
+
+name: Prepare tox
+
+description: Prepare tox-based workflows
+
+runs:
+  using: composite
+  steps:
+    - name: Ensure upgraded pip
+      run: pip install --upgrade pip
+      shell: bash
+    - name: Install tox
+      run: pip install tox
+      shell: bash
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -34,8 +34,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y tar net-tools procps tmux ethtool uml-utilities \
-                                git git-lfs graphviz libenchant-2-dev \
+        sudo apt-get install -y tar net-tools procps uml-utilities \
                                 openvswitch-switch qemu-kvm qemu-utils dnsmasq \
                                 ntfs-3g iproute2 libpcap-dev
     - name: Prepare minimega
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,0 +1,50 @@
+# This workflow will install Python dependencies and run tests with a variety of Python versions
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: Testing
+
+on:
+  push:
+    branches: [ "*" ]
+  pull_request:
+    branches: [ "main" ]
+
+env:
+  LOG_DIR: /var/log/firewheel
+  MINIMEGA_CONFIG: /etc/minimega/minimega.conf
+  # Set the FIREWHEEL environment variables
+  EXPERIMENT_INTERFACE: lo
+  MM_BASE: /tmp/minimega
+  MM_INSTALL_DIR: /opt/minimega
+  GRPC_HOSTNAME: localhost
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y tar net-tools procps tmux ethtool uml-utilities \
+                                git git-lfs graphviz libenchant-2-dev \
+                                openvswitch-switch qemu-kvm qemu-utils dnsmasq \
+                                ntfs-3g iproute2 libpcap-dev
+    - name: Prepare minimega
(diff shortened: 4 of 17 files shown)


.github/actions/prepare-tox/action.yml
```diff
@@ -0,0 +1,15 @@
+# A composite action to prepare tox-based workflows
+
+name: "Prepare tox"
+
+description: "Prepare tox-based workflows"
+
+runs:
+  using: "composite"
+  steps:
```
sdelliot, line 9: It seems reasonable to also require installing Python here. Are there compelling reasons why choosing a standard version (3.11) and using that for all tox jobs wouldn't work?

mitchnegus: The reason I had not included that one was because I hadn't put thought into how the matrixing versions (in lint) and full inclusion of all versions (in the tests) should be reconciled.

When we run `tox -e py39,py310,...` it runs an individual tox job for tests for each. That means that the runner has to have all those versions of Python installed to start with, because tox will look for each Python as it runs each job. But for linting, our `tox -e lint` job will just use whichever Python is the default in our virtual environment (I think). 

The `setup-python` action seems to allow you to [specify the default Python version](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#specifying-multiple-pythonpypy-versions)as the last version identified among multiple selected. I'd think if nothing else we might want to control which version is used for linting (as either the most recent Python version, or the second most recent if we think there's more stability benefits for that).

Also, another concern that popped into my head might be that there could be a performance penalty for installing all versions of Python for the linting job. From what I can see on the pipelines right now, it's [less than 1 second for just one version at a time](https://github.com/sandialabs/firewheel/actions/runs/11865058215/job/33069575968) and [closer to 9 seconds to install all 6 (Python 3.8 through 3.13)](https://github.com/sandialabs/firewheel/actions/runs/11865058209). That's not huge for a CI pipeline—especially if it's running concurrent with tests which will almost always take far longer—so I'm inclined to think it's not a big deal.

mitchnegus: Ok, I've thought more about this, and I think adding Python here would make things difficult. Based on the comment above, in these tests we need to do three distinct things:
1. Unit test – `tox run`/`tox -e py39,py310...`: Run unit tests with tox, with tox running tests for each version of Python. This requires that all tested versions of Python be installed on the runner for this one job.
2. Documentation – `tox -e docs: Build documentation with just any old Python version that is installed (only needs to be done once). 
3. Lint – `tox -e lint`: Run linting with tox, where tox will run the linting job with whatever Python version is the default. Hence, we use the version matrixing so that only the Python version we want to use for linting is installed for a given workflow, and tox uses that (e.g., job 1 installs 3.9 and lints with 3.9, before job 2 installs 3.10 and lints with 3.10, etc.).


That all means that (1) needs all versions installed, (2) only needs one version installed, but if there are multiple versions it should be fixed as some mid-lifecycle default (e.g., 3.11), and (3) only needs one version installed at a time, but must be run multiple times with the majority of those runs being something _other_ than the default.

Reconciling these conditions, especially if we install all the versions but then need to explicitly call tox with one version of Python or the other, is a more difficult means of accomplishing the same effect as installing only the versions we need for each job but then performing the rest of the tox setup in a common manner.

(I'm also still _quite_ green in my understanding of GitHub workflows/actions, so if you think this misses the mark I'm open to other suggestions.)

.github/actions/prepare-discovery/action.yml
```diff
@@ -0,0 +1,15 @@
...
+description: Install discovery
+
+runs:
+  using: composite
+  steps:
+    - name: Install discovery
+      run: |
+        wget https://github.com/mitchnegus/minimega-discovery/releases/download/firewheel-debian_faed761/discovery.deb
+        sudo dpkg -i discovery.deb
+        sudo chown -R $USER:minimega /opt/discovery
```
mitchnegus, line 14: As a note: I did not include dependency installation for either minimega, discovery, or FIREWHEEL in the preparation scripts. I found that each call to `sudo apt update` and `sudo apt install` had enough overhead to increase the test runtime by a decent amount. However, now that the tests are working and I'm not debugging, we may want to consider if that overhead is worth having these composite actions be self-sufficient.

.github/actions/prepare-minimega/action.yml
```diff
@@ -0,0 +1,29 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Install minimega
+      run: |
+        wget https://github.com/sandia-minimega/minimega/releases/download/2.9/minimega-2.9.deb
+        sudo dpkg -i minimega-2.9.deb
+        sudo chown -R $USER:minimega $MM_INSTALL_DIR
+        sudo ln -s $MM_INSTALL_DIR/bin/minimega /usr/local/bin/minimega
+        sudo ln -s $MM_INSTALL_DIR/bin/minimega /usr/local/bin/mm
```
mitchnegus, line 16: As a note: I did not include dependency installation for either minimega, discovery, or FIREWHEEL in the preparation scripts. I found that each call to sudo apt update and sudo apt install had enough overhead to increase the test runtime by a decent amount. However, now that the tests are working and I'm not debugging, we may want to consider if that overhead is worth having these composite actions be self-sufficient.

.github/actions/prepare-discovery/action.yml
```diff
@@ -0,0 +1,15 @@
...
+name: Prepare discovery
+
+description: Install discovery
+
+runs:
+  using: composite
+  steps:
+    - name: Install discovery
+      run: |
+        wget https://github.com/mitchnegus/minimega-discovery/releases/download/firewheel-debian_faed761/discovery.deb
```
mitchnegus, line 12: I created a deb file for discovery based on my best attempts to update the [discovery](https://github.com/sandia-minimega/discovery/tree/master) source code for a modern version of Go. Pending the results of sandia-minimega/discovery#15, we should be able to remove the dependence on my personal fork.

pyproject.toml
```diff
@@ -55,28 +55,30 @@ classifiers = [
...
 [project.optional-dependencies]
-test = [
-    "tox<=4.23.2",
-    "coverage<=7.6.10",
-    "pytest-cov<=6.0.0",
+mcs = [
+    "firewheel-repo-base",
+    "firewheel-repo-linux",
+    "firewheel-repo-vyos",
 ]
```
mitchnegus, line 82: We may wish to add the full set of MCs here (including DNS, NTP, layer 2, and tutorials). I only added these three since they are the minimum set required to run the full test suite (provided the appropriate images are available).

.github/workflows/testing.yml
```diff
@@ -0,0 +1,50 @@
...
+  EXPERIMENT_INTERFACE: lo
+  MM_BASE: /tmp/minimega
+  MM_INSTALL_DIR: /opt/minimega
+  GRPC_HOSTNAME: localhost
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
mitchnegus: I wasn't sure. I thought about it a bit and decided not to because it seemed like it might be good to know the results for each specific Python version.

The recent NetworkX upgrade serves as a good example: when I tested that PR using this test suite, the tests all completed, failing for 3.8 and 3.9 while passing for 3.10 and beyond. It made it really obvious that the error was likely just dependency issues, and made the debugging really easy to check that NetworkX versions including the tested change had dropped support for old Python versions.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,50 @@
...
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y tar net-tools procps tmux ethtool uml-utilities \
+                                git git-lfs graphviz libenchant-2-dev \
+                                openvswitch-switch qemu-kvm qemu-utils dnsmasq \
+                                ntfs-3g iproute2 libpcap-dev
```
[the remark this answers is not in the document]
mitchnegus: Oh true, I hadn't actually thought about those. I stripped out a bunch of the LaTeX ones because they were causing obvious delays, but I agree we can cut out the ones that aren't required. I just copied and pasted from our documentation. (Which at least means the docs are generally pretty good!)

.github/workflows/testing.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Prepare minimega
+      uses: ./.github/actions/prepare-minimega
+    - name: Prepare discovery
+      uses: ./.github/actions/prepare-discovery
+    - name: Prepare FIREWHEEL
+      uses: ./.github/actions/prepare-firewheel
+    - name: Run unit tests
+      run: |
+        firewheel test unit -m 'not long and not mcs' \
+                            --cov --cov-report=term --cov-fail-under=60
```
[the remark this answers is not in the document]
mitchnegus: Whoops, I explained this in the extended body of the commit message because I expected you to ask :grin: (but then forgot to include it and elaborate in these notes). 

The previous test level was not enough to get the unit tests to pass by themselves. Without a lot of digging, I think the reason is twofold:
1. The previous coverage level included functional tests, which we can decide if we'd like to include or not. I don't know if we ever discussed that, but my gut feeling was that we'd want to know coverage for unit tests (i.e., are we testing the unit-level behavior of absolutely everything), while the value of the functional tests is at a higher level (i.e., does this stuff all work together).
2. I've excluded tests that depend on other MCs or which take a while to run, meaning that any lines hit only by those are now uncovered. I think we actually _do_ need to adjust that so that the long-running tests do get run at some point (although maybe we find a way to run them only on merges, not every push).

pyproject.toml
```diff
@@ -55,28 +55,30 @@ classifiers = [
...
     "qemu.qmp==0.0.3",
-    "rich>=13.6.0,<13.10",
     "requests>=2.22.0,<=2.32.3",
-    "importlib_metadata>=3.6,<=8.5.0",
+    "rich>=13.6.0,<13.10",
 ]
 
 [project.optional-dependencies]
-test = [
-    "tox<=4.23.2",
```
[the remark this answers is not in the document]
mitchnegus: I'll probably leave it in dev? We still have the ability for a user to do `tox run` or something like `tox -e py311` which will run tests. This feels like a fair thing to leave as a developer capability (along with all the other linting/typing/etc. things that tox does).

.github/workflows/testing.yml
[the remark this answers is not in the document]
mitchnegus: More than 2 mins down to closer to 30 seconds—that's a win!

.github/workflows/testing.yml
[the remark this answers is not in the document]
sdelliot: All of that reasoning make sense to me. Let's make an issue for the lack of functional tests and a separate one for running the long-running tests.

sdelliot (approved):
LGTM!

PR #14, opened 2024-11-20
Using built-in tools to automatically version FIREWHEEL

sdelliot:
This PR moves away from manually defined versioning to using [setuptools-scm](https://setuptools-scm.readthedocs.io/en/latest/) and our GitHub releases to version the software.


mitchnegus (approved):
Looks correct to me!

PR #21, opened 2024-12-02
Allow workflow reuse across the FIREWHEEL ecosystem

mitchnegus:
This adds the `workflow_call` keyword to the current workflows, enabling the workflows to be used as [reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows). This will allow us to essentially import the standard FIREWHEEL workflows in our model component repositories (see sandialabs/firewheel_repo_base#1).


sdelliot (commented):
Looks good to me.

sdelliot (approved):

PR #26, opened 2024-12-03
Removing outdated references to Elasticsearch

sdelliot:
This PR updates the documentation to fix an outdated comment that Elasticsearch is installed with FIREWHEEL. Additionally, we fix a bug introduced by #14 which neglected to update the version for the Sphinx documentation.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -13,11 +13,11 @@
 import sys
 import importlib.util
 from pathlib import Path
+from importlib.metadata import version as importlib_version
 from importlib.metadata import metadata
 
 import sphinx_rtd_theme  # noqa: F401
 
-import firewheel
 from firewheel.control.repository_db import RepositoryDb
 from firewheel.control.model_component_manager import ModelComponentManager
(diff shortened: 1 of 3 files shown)


sdelliot:
> This looks good to me. I think I missed the fact that in #14 we actually got rid of the `firewheel.__version__` variable. Do you think we might be able to restore that, but using the Importlib version accessor? I haven't used importlib in that way before (I've always opted for using a `_version.py` file) so I'm not sure what it's limitations are as far as being callable in the `firewheel.__init__` module.
> 
> My thinking is just that I will often try to call `<package>.__version__` on packages, especially when debugging, and so it's a nice convenience to have around. I'm happy to merge this PR as is and investigate ways to bring it back if need be. I don't want to hold this one up.

It looks like based on the latest packaging guide: https://packaging.python.org/en/latest/discussions/versioning/#accessing-version-information-at-runtime it seems like it is not required, but many projects choose to do so. Given that there is minimal overhead to maintain. I'll add that back but use the `importlib.metadata` method to generate the version number.

PR #38, opened 2025-01-06
Add progress bars for experiment launch timing

mitchnegus:
This builds off a bunch of previous work, but adds a handful of new features too. The big ideas are:

### Experiment Setup Progress Bars

This adds progress bars for graph creation and experiment setup; these use combine progress bars and rich `Live` displays so that output are presented cleanly (e.g., graph creation progress, VM cache progress, and MC printouts do not clobber each other).

Notably, this MR extends the `AbstractPlugin` to accept an (optional) `console` keyword argument. This should have no impact on existing MC plugins that do not display output, but it allows MCs that do display output during setup to be able to avoid competition with any live output coming directly from FIREWHEEL.

### Negative Time Progress Bar

The MR also adds a progress bar for experiment negative time when running `firewheel vm mix`. While defining progress through negative time is only particularly well-defined on a per VM basis, it seemed less than ideal to show _n_ progress bars for each of the _n_ VMs in the experiment. Instead, this uses the heuristic that overall progress through negative time is defined as the number of (negative time) schedule events that have been completed per the total number of negative time schedule events across all VMs. 

For practicality, "event" completion is defined for each VM as any negative time event that was scheduled before the VMs current time, which means that this heuristic is strictly conservative. Some subset of events in the queue that were all scheduled at the same time may actually have completed, but since the VM time has not advanced, they are not counted as "complete" by this metric.

Noting that schedules can be adapted as the experiment evolves, the total number of negative time events is not guaranteed to remain constant. Rich allows progress bars to update their total values, so this does not cause any trouble in implementation; however; it might surprise users who notice that their totals aren't fixed. 

Finally, when the experiment is setting up or in positive time, the negative time progress bar is hidden to avoid visual clutter.


src/firewheel/control/experiment_graph.py
```diff
@@ -25,16 +25,19 @@ class AbstractPlugin:
...
 
         Args:
             graph (ExperimentGraph): The experiment graph in which the plugin will execute.
             log (logging.Logger): A log the plugin can use.
+            console (rich.console.Console): An optional rich console to use
+                for the MC output.
         """
         self.g = graph  # pylint: disable=invalid-name
         self.log = log
+        self.console = console
```
sdelliot, line 40: This will be super useful! It's a nice/simple feature to have.

sdelliot (commented):
Generally, I like both these ideas. I tested this out and everything seemed to work as expected. However, this is a large PR with two distinct (although related) features. Can you please split them up into their distinct components? I'll go back through and review the code after that has happened.

PR #47, opened 2025-01-21
Addressing a deprecated assumption from NetworkX "links" data. S…

sdelliot:
The "link" keyword is deprecated for the `node_link_data` function and will be removed in version  NetworkX v3.6. We are switching to the recommended default "edges" instead.

See [node-link-data](https://networkx.org/documentation/stable/reference/readwrite/generated/networkx.readwrite.json_graph.node_link_data.html#node-link-data) for more details.

### DIFF
--- a/src/firewheel/control/dependency_graph.py
+++ b/src/firewheel/control/dependency_graph.py
@@ -219,7 +219,7 @@ class DependencyGraph:
                     {"id": <id>, "type": <"entity" or "constraint">},
                     ...
                 ],
-                "links": [
+                "edges": [
                     {"source": <id>, "target": <id>},
                     ...
                 ],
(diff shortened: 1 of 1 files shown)


mitchnegus:
It appears that this change was introduced in NetworkX 3.4 which [drops support for Python versions less than 3.10](https://github.com/networkx/networkx/blob/fe7795acbc31518bb5154de79afb0b9dab2f4846/pyproject.toml#L9). We support 3.8 and 3.9, so it looks to me like this will break any installations using those versions.

Otherwise, it's 👍🏼 . There are some references to those same values in the tests that I updated, and so the [tests now pass for versions that comply with the NetworkX version restrictions](https://github.com/mitchnegus/firewheel/actions/runs/12995144836).

sdelliot:
Excellent point! NetworkX follows: https://scientific-python.org/specs/spec-0000/. I agree that we should hold off on this until we drop support for Python 3.9.

PR #48, opened 2025-01-23
refactor: Replace the gRPC-backed RepositoryDB with a JSON-file backed one.

sdelliot:
This is an effort to reduce the impact of our services on starting and using FIREWHEEL. Specifically, the RepositoryDB has been part of the gRPC service that FIREWHEEL provides. This inter-dependency is unnecessary for the following reasons:

1. It does not need to be accessed except from the "control node".
2. The gRPC service stored the repositories (e.g., a path) in a JSON file.
3. As repositories become pip packages, the necessity of this feature wains entirely.

In this PR, we strip away the unnecessary complexity of gRPC and simply rely on the single JSON file for locally-installed model components. We have updated all necessary documentation and tests accordingly. Finally, we needed to provide a minimum version for the `grpcio` package as the newly generated code checks version information.

Note that the `pb2*` files are auto-generated via `grpc_tools.protoc`.


src/firewheel/control/repository_db.py
```diff
@@ -23,28 +26,22 @@ class RepositoryDb:
...
         self,
-        host=config["grpc"]["hostname"],
-        port=config["grpc"]["port"],
-        db=config["grpc"]["db"],
+        db_basepath=config["system"]["default_output_dir"],
+        db_filename="repositories.json",
     ):
         """
-        Set up the connection to the gRPC server.
+        Set up the class variables and path to the RepositoryDB file.
```
mitchnegus, line 33:
```suggestion
        Set up the instance variables and path to the RepositoryDB file.
```

docs/source/_static/proto_doc.html
```diff
@@ -63,57 +63,57 @@
       }
 
       td p:nth-child(1) {
-        text-indent: 0;
+        text-indent: 0;  
       }
 
-
-      .field-table td:nth-child(1) {
+
```
mitchnegus, line 69: Lot's of extraneous whitespace got added to this file. I assume it's also autogenerated, but if not might be worth stripping it back out.

sdelliot: It's definitely autogenerated. I minified it in [652bab0](https://github.com/sandialabs/firewheel/pull/48/commits/652bab0df5868954810a423a151fa24922ab9511), which greatly reduced filesize!

mitchnegus (approved):
@sdelliot Approving this because the code all looks good to me. We should probably look into why the PR label checker is failing. My quick read is that it can't find the branch `minimize-hard-dependencies` because that branch doesn't exist on this base repo, it only exists on your fork. Not sure if there's a way to fix that via configuration?

PR #49, opened 2025-01-23
Feature: restructure SSH

sdelliot:
Like #48, this PR serves to increase modularity in FIREWHEEL and enable a subset of FIREWHEEL use with minimal "hard" dependencies. In this case, the `SSHManager` (and related functions) ultimately rely on minimega. However, minimega is **NOT** required for the other CLI utilities. However, due the single module, minimega was an implicit requirement for `utils.py`, and subsequently the entire CLI. This MR explicitly moves the SSH functionality to a new module which will enable the CLI to function without a functioning minimega installation.

With the exception of a single change, this is a straightforward moving of functionality. The one change was adding ``("-o", "HostKeyAlgorithms=+ssh-rsa"),`` as a default SSH option. The reason for this addition is that newer operating systems (e.g., Ubuntu 22.04) have stopped supporting this algorithm by default, but many of the images needed for experiments (older OS versions) will use this algorithm by default, preventing SSH access. This can be removed if requested.


mitchnegus (approved):
This all looks good to me now.

PR #56, opened 2025-01-28
Update documentation with minor changes

mitchnegus:
This provides a few minor updates to the docs.

1. I clarify that the `PYTHON_BIN` environment variable is the executable (which may be either a path or a symlink). I think this wording might be better, because just giving `python3` (assuming that it's being run in a virtual environment) doesn't strike me as a path.
2. It appears that `libenchant-dev` has been superseded by `libenchant-2-dev` and is accessible with that new name in the relevant Ubuntu repositories. (Similar for Fedora-based OSs and `enchant2-devel` and `enchant-devel`.)

### DIFF
--- a/docs/source/install/prereqs.rst
+++ b/docs/source/install/prereqs.rst
@@ -25,7 +25,7 @@ The following command will install all required system packages for FIREWHEEL an
 
 For users who are improving FIREWHEEL, building documentation, or running the FIREWHEEL tests, these packages are also needed::
 
-    sudo apt-get install -y graphviz texlive-latex-recommended texlive-fonts-recommended texlive-latex-extra latexmk libenchant-dev
+    sudo apt-get install -y graphviz texlive-latex-recommended texlive-fonts-recommended texlive-latex-extra latexmk libenchant-2-dev
 
 CentOS
 ------
--- a/docs/source/install/prereqs.rst
+++ b/docs/source/install/prereqs.rst
@@ -13,7 +13,7 @@ System Dependencies
    :start-after: dependencies-inclusion-marker
    :end-before: dependencies-inclusion-stop
 
-This section quickly outlines the system packages needed for full FIREWHEEL functionality with the exception of Python, discovery, minimega, git, and git-lfs as those are discussed in detail below.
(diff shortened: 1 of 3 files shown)


sdelliot (changes requested):
It looks like `libenchant-2-dev` is available for Ubuntu versions >=20.04. We do need to update our documentation about supported/tested Operating Systems anyways (see #57). I am flexible as to whether that issue is resolve here or not, but I do think that we should mention that the install strings are specifically for Ubuntu 20.04 and RHEL 9 to prevent confusion for those running older operating systems.

mitchnegus:
Ok, so how about I add a note about which OS are used for the stated dependencies (Ubuntu 22.04 and RHEL 9) and I'll leave #57 since we already have an issue for that one. That'll keep this one really quick which was the intention (and keep that one as a good first issue).

sdelliot:
> Ok, so how about I add a note about which OS are used for the stated dependencies (Ubuntu 22.04 and RHEL 9) and I'll leave #57 since we already have an issue for that one. That'll keep this one really quick which was the intention (and keep that one as a good first issue).

That would be perfect!

docs/source/install/prereqs.rst
```diff
@@ -13,7 +13,7 @@ System Dependencies
    :start-after: dependencies-inclusion-marker
    :end-before: dependencies-inclusion-stop
 
-This section quickly outlines the system packages needed for full FIREWHEEL functionality with the exception of Python, discovery, minimega, git, and git-lfs as those are discussed in detail below.
+Below is a brief outline of the system packages needed for full FIREWHEEL functionality with the exception of Python, discovery, minimega, git, and git-lfs as those are discussed in further detail in subsequent sections.
```
mitchnegus, line 16: I reworded this phrasing because the way our docs display, the outline is "below" and the extended discussion of additional packages are on other (linked) pages.

mitchnegus:
I've update the documentation based on the suggestions, and I think this is ready to go unless you see other things or want to update anything further.

sdelliot (approved):
This all looks good to me.

PR #61, opened 2025-02-05
Add a get method to the Vertex object to simulate a dictionary

mitchnegus:
At least one model component [assumes that the `Vertex` object fully implements dictionary behavior](https://github.com/sandialabs/firewheel_repo_dns/blob/2acb4db042f219141b6a5947258124fff670214e/src/firewheel_repo_dns/populate_zones/plugin.py#L57) (almost certainly due to suggestions I made). 

This adds the `get` method to the `Vertex` object to provide that functionality.

### DIFF
--- a/src/firewheel/control/experiment_graph.py
+++ b/src/firewheel/control/experiment_graph.py
@@ -721,6 +721,27 @@ class Vertex(ExperimentGraphDecorable):
         if name:
             self.name = name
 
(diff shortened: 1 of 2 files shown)


mitchnegus:
Closing, because the referenced MC behavior is technically invalid, and a `get` method is not actually required.

PR #65, opened 2025-02-17
fix: Address adding duplicate repositories and conflicts with manual vs package installed repositories.

sdelliot:
This PR fixes two bugs:
1. An issue where multiple same-path repositories could be added.
2. Test cases which failed if there were package installed repositories (in addition to manually installed ones).

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -47,7 +47,7 @@ jobs:
       checks: write
     runs-on: ubuntu-latest
     steps:
-      - uses: danielchabr/pr-labels-checker@v3.1
+      - uses: docker://agilepathway/pull-request-label-checker:latest
         with:
-          hasSome: feature,fix,style,changed,refactor,perf,test,build,ci,chore,revert,deprecated,removed,security,documentation,dependencies
-          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          any_of: feature,fix,style,changed,refactor,perf,test,build,ci,chore,revert,deprecated,removed,security,documentation,dependencies
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release-drafter.yml
```diff
@@ -47,7 +47,7 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: danielchabr/pr-labels-checker@v3.1
+      - uses: docker://agilepathway/pull-request-label-checker:latest
         with:
-          hasSome: feature,fix,style,changed,refactor,perf,test,build,ci,chore,revert,deprecated,removed,security,documentation,dependencies
-          githubToken: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+          any_of: feature,fix,style,changed,refactor,perf,test,build,ci,chore,revert,deprecated,removed,security,documentation,dependencies
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
```
sdelliot, line 53: I am no longer sure if this change is necessary.

.github/workflows/release-drafter.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request:
+  pull_request_target:
```
sdelliot, line 7: This also comes with some [security concerns](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) and we should evaluate carefully prior to merging.

.github/workflows/release-drafter.yml
sdelliot: I think this should be punted to #69 and removed from this PR.

mitchnegus (commented):
This looks good and works for me. I'm still finding that Git cloned MCs will conflict with pip installed editable MCs. 

It looks like under this scheme those are still treated as separate because during comparison one might have a `repository["path"]` like `/opt/firewheel/model_components` while another has a path like `/opt/firewheel/model_components/firewheel_repo_base/src/firewheel_repo_base`.

mitchnegus (approved):

PR #66, opened 2025-02-19
ci: Enhance CI reusability

mitchnegus:
CI pipelines for model component repos currently cannot access the composite actions that are called in the reusable workflows for this repo. It seems that this is because [GitHub actions defined in a called repo](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#overview) (e.g., `sandialabs/firewheel`) are executed as if they were defined in the caller repo (e.g., the MC repo).

An example of this failing behavior can be [found in the recent CI pipeline](https://github.com/sandialabs/firewheel_repo_dns/actions/runs/13420030343/job/37490153500#step:4:1) for the `firewheel_repo_dns` repo. That repo cannot find `.github/actions/prepare-tox` because it's looking for it in the `sandialabs/firewheel_repo_dns` repository rather than `sandialabs/firewheel` where it actually lives.

I was hoping there might be a way to tell the caller repo to look for things in the called repo, but the closest related info I've found so far is [an issue suggesting this might not be possible](https://github.com/orgs/community/discussions/66094). There are many suggestions there, including those for choosing specific branches, but I think that it probably just makes the most sense for us to use the full path to the action (e.g., `sandialabs/firewheel/.github/actions/prepare-tox`) in our workflows until we need a more sophisticated system.

I quote the link to that issue here as a place to go looking if we eventually need a more complete strategy, such as a new repo only containing actions and similar to the CodeQL Action repo that is linked there.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -37,7 +37,7 @@ jobs:
       with:
         python-version: '3.x'
     - name: Prepare to use tox-based environments
-      uses: ./.github/actions/prepare-tox
+      uses: sandialabs/firewheel/.github/actions/prepare-tox
     - name: Build Documentation
       run: |
         tox -e dependencies,docs
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -27,7 +27,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Prepare to use tox-based environments
-      uses: ./.github/actions/prepare-tox
+      uses: sandialabs/firewheel/.github/actions/prepare-tox
     - name: Lint code
       run: |
         tox -e lint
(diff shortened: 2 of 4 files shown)


mitchnegus:
Also, that `danielchabr/pr-labels-check` job seems to be unable to identify branches on forks. @sdelliot Do you know if there's a way to configure it to look at forked branches?

sdelliot:
> Also, that `danielchabr/pr-labels-check` job seems to be unable to identify branches on forks. @sdelliot Do you know if there's a way to configure it to look at forked branches?

I believe that this is because we are not using [`pull_request_target`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target) in the `.github/workflows/release-drafter.yml` (see https://github.com/sandialabs/firewheel/pull/65/files) but it seems hard to test this relies on using information from the primary repository so merging/testing seems to be the only way. Additionally, there are some known [security concerns](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) that we will want to evaluate prior to making that change.

sdelliot (approved):
LGTM

PR #68, opened 2025-02-20
feat: Create a initial docker container for FIREWHEEL

sdelliot:
This PR provides a working Docker container for FIREWHEEL. We should consider this as a beta-testing phase and remove that label in a future release. This container has been tested in both low-trust and high-trust environments where various permissions/kernel modules were stripped and verified functionality. The associated documentation provides a good overview of the container and how it was created.
Aside from that, there are a few important things to note:
- The FIREWHEEL docker container is based on the [minimega](https://github.com/sandia-minimega/minimega/) container image.
- Multi-node/containers have not been tested/explored.
- This PR does not include a Docker compose environment. This should happen in a different PR.
- The container does not contain any VM images and those will need to be added post-launch by users.


docker/fsroot/start-minimega.sh
```diff
@@ -0,0 +1,54 @@
...
+/usr/share/openvswitch/scripts/ovs-ctl start 1>/var/log/minimega.log 2>/var/log/minimega.log
+if [ $? -ne 0 ]; then
+    echo "Failed to start Open vSwitch" 1> /var/log/minimega.log
+fi
+
+# Check if Minimega is already running
+if pgrep -f "/opt/minimega/bin/minimega" > /dev/null; then
+    echo "Minimega is already running. Exiting script." 1> /var/log/minimega.log
+    exit 0
+fi
```
sdelliot, line 12: @GhostofGoes and @jacdavi do you think that minimega would benefit from these additions to the `start-minimega.sh` or do you think it would cause a loss of functionality?

docker/firewheel.dockerfile
```diff
@@ -0,0 +1,105 @@
...
+    echo "LANG=\"en_US.UTF-8\"" > /etc/locale.conf || { echo "Locale setup failed"; exit 1; }
+### Locale Support END ###
+
+
+# --#--#--#--#--#--#--#--#--#--#--#--#--#--#--#--#--#--#--#--#--#--#--#--#-- #
+
+WORKDIR /
+
+# Install discovery
+RUN wget https://github.com/mitchnegus/minimega-discovery/releases/download/firewheel-debian_faed761/discovery.deb && \
```
gregjacobus, line 38: Is this where we want to be pulling from?

sdelliot: Eventually, we want to pull from https://github.com/sandia-minimega/discovery/ but a new release hasn't been issued yet. I plan on pinging them on this offline.

docker/fsroot/start-minimega.sh
sdelliot: The primary logic change is exiting the when minimega is already running. We call this script to start minimega when clearing our testbed, but if minimega (and miniweb) are still running, that isn't a useful action. Especially as trying to start multiple instances of minimega/miniweb seemed to cause issues (even with `-force True`).

docker/firewheel.dockerfile
```diff
@@ -0,0 +1,105 @@
+FROM ghcr.io/sandia-minimega/minimega/minimega:master AS minimega
```
sdelliot, line 1:
```suggestion
FROM ghcr.io/sandia-minimega/minimega:master AS minimega
```

This path was updated as of https://github.com/sandia-minimega/minimega/pull/1552

PR #69, opened 2025-02-21
ci: Updating CI to use specific tagged versions of github actions and minimizing/enforcing permissions for all actions.

sdelliot:
This set of changes should enable a more secure CI pipeline (e.g., avoiding supply chain attacks) while still maintaining the current functionality. I suspect that the pipeline will still fail prior to merge due to some of the restrictions. Any issues after that can be addressed quickly.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -11,6 +11,9 @@ on:
   # Allow reuse across the FIREWHEEL ecosystem
   workflow_call:
 
+permissions:
+  contents: read
+
 jobs:
   lint:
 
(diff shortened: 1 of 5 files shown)


mitchnegus (approved):
Looks good here. 

The `pull_request_target` tag is only used here in jobs that do not modify/touch the source code (just GitHub interface level actions).

mitchnegus:
Useful additional context:

- [`pull_request_target`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target) docs
- GitHub Security Lab blog post: [Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/)
- [Stack Overflow explanation](https://stackoverflow.com/a/74959635)

PR #70, opened 2025-02-21
fix: Making adjustments to the tab completion to work with `/bin/sh`

### LINKED ISSUE #67: Tab completion does not work when the default shell does not incorporate "source"
Tab completion does not work when the default shell does not incorporate "source". For example, if the default `$SHELL` is `/bin/sh` than a user may get the following output:
```bash
$ firewheel exp<tab>/bin/sh: 2: [[: not found
/bin/sh: 3: source: not found
```

sdelliot:
This PR enables shells that do not have `source` or the ability to use `[[` such as `/bin/sh`. Closes #67.

### DIFF
--- a/src/firewheel/cli/completion/completion-template.sh
+++ b/src/firewheel/cli/completion/completion-template.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 # This script provides tab completion for FIREWHEEL from Bash or ZSH (non-interactive mode)
 # It iterates over the FIREWHEEL Helpers and adds them to the completion list.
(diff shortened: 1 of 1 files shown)


mitchnegus (approved):
Looks good to me. All seems fine.

mitchnegus:
Note for the future: This fix is necessary because the `$SHELL` variable used for executing snippets within the script pulls from the user's default `SHELL` variable. This may be different than the shell running the script—e.g., even if the script is run via `/bin/bash` as per the hashbang, if the user has set (implicitly or explicitly) `SHELL=/bin/sh` then the sourcing will happen with that shell instead.

PR #76, opened 2025-03-03
fix: Correct behavior of the SSH Manager for quoted strings

mitchnegus:
@gregjacobus was running into an issue where the SSH manager would fail to execute a `grep` command when it was passed via the command line. It appears that the issue stemmed from a failure to use `shlex` when recombining arguments parsed from the command line.

This updates the SSH manager to ensure that `shlex` is used for all parsing and recombining operations when manipulating arguments passed to as commands to the SSH/SCP commands. 

The PR also adds a few extra test cases to have better coverage of quoted commands.


sdelliot (approved):
LGTM

PR #78, opened 2025-03-07
fix: Fixing hard-coded namespace

gregjacobus:
Don't think this actually fixes the cluster/namespace issues we were having, but this seems useful regardless

### DIFF
--- a/src/firewheel/cli/helpers/start
+++ b/src/firewheel/cli/helpers/start
@@ -66,6 +66,18 @@ fi
 
 DONE
 
+
+RUN Shell ON compute
+MM_INSTALL_DIR="$("$FIREWHEEL_PYTHON" "$FIREWHEEL" config get minimega.install_dir)"
+MM_BASE_DIR="$("$FIREWHEEL_PYTHON" "$FIREWHEEL" config get minimega.base_dir)"
+MINIMEGA_BIN="$MM_INSTALL_DIR/bin/minimega -base=$MM_BASE_DIR"
+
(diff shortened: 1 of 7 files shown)


src/firewheel/cli/helpers/start
```diff
@@ -66,6 +66,18 @@ fi
...
 
+
+RUN Shell ON compute
+MM_INSTALL_DIR="$("$FIREWHEEL_PYTHON" "$FIREWHEEL" config get minimega.install_dir)"
+MM_BASE_DIR="$("$FIREWHEEL_PYTHON" "$FIREWHEEL" config get minimega.base_dir)"
+MINIMEGA_BIN="$MM_INSTALL_DIR/bin/minimega -base=$MM_BASE_DIR"
+
+# Ensure the specified namespace is created and that localhost is added
+$MINIMEGA_BIN -e namespace "$(firewheel config get minimega.namespace)"
+$MINIMEGA_BIN -e ns add-hosts localhost
```
sdelliot, line 77: What is the benefit of adding localhost? Does this cause conflicts when the server is added to the namespace?

gregjacobus: localhost isn't added by default to the firewheel namespace (only to the default minimega one), so we have to add it ourselves. We add other servers currently by "mesh dial", but you can't mesh dial yourself. I'm open to restructuring so every host gets added the same way. This should probably still be marked as a draft though, i don't feel super confident yet in all of this

PR #81, opened 2025-03-14
fix: Ensuring users know when a duplicate repository is trying to be added

sdelliot:
This PR further enhances #65 by ensuring the user gets feedback on if duplicate repositories are installed. Additionally, it performs more test clean up.


mitchnegus:
@sdelliot Do you know how this differs from #80? The code looks good on my first run through, but I'd like to be sure I'm solid about what I'm looking for if I test this out.

gregjacobus:
Maybe this is overly picky, so feel free to ignore. But I don't love wrapping the entire _validate_repository function in the try/except, especially catching a ValueError that could accidentally mask real exceptions in the future. Thoughts on:

1) Making the ValueError a FileExistsError (although this feels a tad misleading)
2) Creating a simple custom RepositoryAlreadyExistsException

mitchnegus:
I would support that, I also don't like overbroad exceptions. I like the idea of a `RepositoryAlreadyExistsError` or `RepositoryExistsError` (maybe a bit shorter?), maybe by subclassing `ValueError`? We could subclass `FileExistsError`, but like you said, I think that's a tad misleading since this isn't really a subclass of `OSError` like `FileExistsError` is.

sdelliot:
Okay, I adjusted this PR and it should be good to go now.

src/firewheel/tests/unit/control/test_repository_db.py
```diff
@@ -25,6 +25,8 @@ def repository_db():
     repository_db = RepositoryDb(
         db_filename="test_repositories.json",
     )
+    for repo in repository_db.list_repositories():
+        repository_db.delete_repository(repo)
     yield repository_db
```
mitchnegus, line 30: Minor suggestion (as discussed), potentially we move the cleanup after the yield to clearly identify this as a "cleanup" action. More information can be found in the [pytest docs](https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization) about this process of "fixture finalization".

```suggestion
    yield repository_db
    # Ensure all repositories are removed during teardown
    for repo in repository_db.list_repositories():
        repository_db.delete_repository(repo)
```

mitchnegus (dismissed):
PR looks good. In addition to the suggested change (optional), it might also be worth adding checks for the new return values (integers) in the tests. 

Looks like that'd be:
1. An `assert repository_db.add_repository(repo_entry) == 1` in the `test_add_repository` function; and
2. An `assert repository_db.add_repository(repo_entry) == 1` followed by an `assert repository_db.add_repository(repo_entry) == 0` in the `test_duplicate_repository` function.

(Doesn't seem like I can add a suggestion for those since they're outside the current scope of changes.)

mitchnegus (approved):

PR #94, opened 2025-03-26
ci: Run PR label check only in pull request contexts

mitchnegus:
This is intended to only run the PR label checker for events occuring in a pull request context (cases where the `event_name` is `pull_request_target` and not pushes).


sdelliot (approved):
LGTM

PR #102, opened 2025-04-01
test: Use mock for simulating permissioned systems

mitchnegus:
The `add_repository_invalid` test can fail on systems where the user has read permissions to the `/root` directory. (This is likely uncommon, but not infeasible.) 

This adds a mocking function to simulate the `os.access` function such that it _always_ returns `False` for the `/root` directory. I figure this is preferable than trying to find a file that will not be user-readable on every system, especially because our unit tests have no need to test the behavior of the `os.access` function itself.


sdelliot (approved):
LGTM

PR #104, opened 2025-04-02
ci: Fix pydocstyle version to prevent linting errors

### LINKED ISSUE #103: Updates to (unpinned) pydocstyle break linter
The current pipeline has broken due to an update to pydocstyle. We did not previously pin a pydocstyle version in [`tox.ini`](https://github.com/sandialabs/firewheel/blob/986cb66cbfcd2d6adadbb1fa19b9eced838462aa/tox.ini#L38).

A new command line option ([`--ignore-private-args`](https://jsh9.github.io/pydoclint/config_options.html#14---ignore-private-args-shortform--ipa-default-false)) was [added in pydocstyle 0.6.3](https://github.com/jsh9/pydoclint/blob/main/CHANGELOG.md#063---2025-03-30). While I wouldn't have expected this argument to have affected us based on the documentation description—the default value is `False`; i.e., private arguments _do_ need to appear in the docstring, and ours do—here we are.

I haven't had time to get my mind around if there's an easy fix to this: I don't know that we want to use the new argument to affirmatively ignore private arguments (since we explicitly include them), and I haven't yet figured out why if we're _not_ ignoring private arguments, pydocstyle is saying that our docstrings contain more arguments than the function signature (maybe a bug in the recent pydocstyle update, or maybe I'm just not interpreting things right).

Perhaps we pin to pydocstyle 0.6.2 at first, and then choose if/when to investigate further. If this is a bug, then maybe it resolves itself.

mitchnegus:
This PR fixes the pipeline, currently broken after an update to pydocstyle. We did not previously pin a pydocstyle version in [`tox.ini`](https://github.com/sandialabs/firewheel/blob/986cb66cbfcd2d6adadbb1fa19b9eced838462aa/tox.ini#L38).

A new command line option ([`--ignore-private-args`](https://jsh9.github.io/pydoclint/config_options.html#14---ignore-private-args-shortform--ipa-default-false)) was [added in pydocstyle 0.6.3](https://github.com/jsh9/pydoclint/blob/main/CHANGELOG.md#063---2025-03-30). The default value of that argument—which I believe we should be defaulting to—is `False` (e.g., _do not_ ignore private arguments); however, it seems like even though we are specifying private arguments in our docstrings, the check is still failing.

This PR does not solve the problem or advocate any solution, it simply pins the previous behavior to fix the pipeline while we determine the best course of action.

Closes #103.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ basepython = python3
 skip_install = true
 deps =
     flake8~=7.0.0
-    pydoclint[flake8]
+    pydoclint[flake8]==0.6.2
 commands =
     flake8 {toxinidir}/src/firewheel/ {posargs}
 


sdelliot (approved):
LGTM

PR #113, opened 2025-04-18
feat: Support for using Ansible playbooks for installing model components

sdelliot:
## Current State
Currently, some Model Components may require additional data to be downloaded (e.g., VM resources, etc.) that cannot or should not be stored within the MC repo package. In this case, the Model Components can have an `INSTALL` file, which can be any executable script (as denoted by a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) line). See: [Model Component INSTALL file](https://sandialabs.github.io/firewheel/model_component/mc_install.html#mc-install).

This PR enables using [Ansible](https://ansible.readthedocs.io/) playbooks in lieu of using a bash-based script.

## Primary Benefits
There are several primary motivators for this change:
1. [**Idempotence**](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_intro.html#id6) - By design, it is significantly easier to design Ansible playbooks that can be run multiple times without changing the final state. Specifically, this is necessary when installation fails and needs to be re-run. While other scripts can have this property, Ansible makes it significantly easier to do so.
2. **Readability** - `INSTALL` files that are written with Shell scripts are significantly harder to read/understand than Ansible-based YAML. This not only impacts maintainability, but also the security as reviewing shell code for issues is more difficult. 
3. **Pulling Cached Files** - Downloading files and creating VM Resources or images each time FIREWHEEL is installed is time consuming and, in the case of offline installations, impractical. The common solution for this is to perform this action once and cache the resulting files. By using Ansible, FIREWHEEL can provide a few common ways to retrieve these cached files. When combined with the Idempotence of Ansible scripts, this will further reduce installation errors and speed new installs.

## Key Changes
1. A new top-level FIREWHEEL config option called `ansible` where there is a subkey `cache_type="online"`. That is, by default assume that the system is online and will not pull from any kind of cached system. This top-level option will enable users to add other values (URLs, API Keys, etc.) for getting the various cached files.
2. New Ansible playbooks that support caching from Amazon S3 buckets and from Git repositories.
3. Modifications to the `model_component_install.py` to check and see if the `INSTALL` script is a valid Ansible playbook. If it is, then we will use [ansible-runner](https://pypi.org/project/ansible-runner/) to execute the script. Otherwise, the existing behavior is maintained.

## Pending Tasks

- [ ] The relevant documentation needs to be updated to describe these options.
- [ ] Test cases for the new install code would be valuable, though it may be difficult to construct.
- [ ] Additional cache options should be supported. Options include: JFrog Artifactory, a standard [get_url](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html), MS Onedrive,  etc. These can also be added in future PRs as need arises.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ classifiers = [
     "Topic :: System :: Operating System",
 ]
 dependencies = [
+    "ansible-runner>=2.0.0,<=2.4.1",
     "minimega==2.9",
     "ClusterShell<=1.9.3",
     "colorama<=0.4.6",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,7 +97,7 @@ docs = [
     "sphinx-design",
 ]
 dev = [
-    "firewheel[mcs,format,docs]",
+    "firewheel[format,docs]",
     "pre-commit",
     "tox~=4.0",
 ]
--- a/src/firewheel/control/ansible_playbooks/git.yml
+++ b/src/firewheel/control/ansible_playbooks/git.yml
@@ -30,8 +30,9 @@
 
     - name: Move cached files to the desired location
       ansible.builtin.copy:
-        src: /tmp/fwcache/{{ item.source }}
-        dest: {{ item.destination }}
+        src: "/tmp/fwcache/{{ item.source }}"
+        dest: "{{ item.destination }}"
+        mode: 0666
       loop: "{{ cached_files }}"
 
     - name: Remove the Git repository
--- a/src/firewheel/control/ansible_playbooks/git.yml
+++ b/src/firewheel/control/ansible_playbooks/git.yml
@@ -1,38 +1,40 @@
 ---
 # This playbook will clone a Git repository and move specified cached files
-# The user will be prompted for all relevant Git information including:
-# - The repository URL
-# - The branch name (optional)
-# - The access token (if required for authentication)
+# If you are using an access token, you can specify it in the `git_server` URL
+# For example: https://<token>@github.com/user/repo.git
+# The user will need to provide all relevant information including:
+# - ``git_server`` - The URL of the git server
+# - ``git_repo_path`` - The path to the repo from the server
+# - ``git_branch`` - The branch name (optional); default: main
 
 - name: Clone Git repository and move files
   hosts: localhost
   vars_prompt:
     - name: git_server
       prompt: "Please enter the Git server URL or set `ansible.git_server` in the FIREWHEEL configuration."
-      private: no
-    
+      private: false
+
     - name: git_repo_path
       prompt: "Please enter the path to the git repository or set `ansible.git_repo_path` in the FIREWHEEL configuration."
-      private: no
+      private: false
 
   tasks:
     - name: Clone the Git repository
-      git:
+      ansible.builtin.git:
         repo: "{{ git_server }}/{{ git_repo_path }}"
         dest: "/tmp/fwcache"
         version: "{{ git_branch | default('main') }}"
-        clone: yes
-        update: yes
-        # If using an access token, you can specify it in the URL
-        # For example: https://<token>@github.com/user/repo.git
+        clone: true
+        update: true
       register: git_clone_result
 
     - name: Move cached files to the desired location
-      command: mv /tmp/fwcache/{{ item.source }} ./{{ item.destination }}
+      ansible.builtin.copy:
+        src: /tmp/fwcache/{{ item.source }}
+        dest: {{ item.destination }}
       loop: "{{ cached_files }}"
 
     - name: Remove the Git repository
-      file:
+      ansible.builtin.file:
         path: "/tmp/fwcache"
         state: absent
--- a/src/firewheel/control/ansible_playbooks/git.yml
+++ b/src/firewheel/control/ansible_playbooks/git.yml
@@ -9,11 +9,11 @@
   hosts: localhost
   vars_prompt:
     - name: git_server
-      prompt: "Please enter the Git server URL"
+      prompt: "Please enter the Git server URL or set `ansible.git_server` in the FIREWHEEL configuration."
       private: no
     
     - name: git_repo_path
-      prompt: "Please enter the path to the git repository"
+      prompt: "Please enter the path to the git repository or set `ansible.git_repo_path` in the FIREWHEEL configuration."
       private: no
 
   tasks:
@@ -31,7 +31,6 @@
     - name: Move cached files to the desired location
       command: mv /tmp/fwcache/{{ item.source }} ./{{ item.destination }}
       loop: "{{ cached_files }}"
-      when: git_clone_result.changed
 
     - name: Remove the Git repository
       file:
--- a/src/firewheel/control/ansible_playbooks/git.yml
+++ b/src/firewheel/control/ansible_playbooks/git.yml
@@ -1,0 +1,39 @@
+---
+# This playbook will clone a Git repository and move specified cached files
+# The user will be prompted for all relevant Git information including:
+# - The repository URL
+# - The branch name (optional)
+# - The access token (if required for authentication)
+
+- name: Clone Git repository and move files
+  hosts: localhost
+  vars_prompt:
+    - name: git_server
+      prompt: "Please enter the Git server URL"
+      private: no
+    
+    - name: git_repo_path
+      prompt: "Please enter the path to the git repository"
+      private: no
+
+  tasks:
+    - name: Clone the Git repository
+      git:
+        repo: "{{ git_server }}/{{ git_repo_path }}"
+        dest: "/tmp/fwcache"
+        version: "{{ git_branch | default('main') }}"
+        clone: yes
+        update: yes
+        # If using an access token, you can specify it in the URL
+        # For example: https://<token>@github.com/user/repo.git
+      register: git_clone_result
+
+    - name: Move cached files to the desired location
+      command: mv /tmp/fwcache/{{ item.source }} ./{{ item.destination }}
+      loop: "{{ cached_files }}"
+      when: git_clone_result.changed
+
+    - name: Remove the Git repository
+      file:
+        path: "/tmp/fwcache"
+        state: absent
--- a/src/firewheel/control/model_component_install.py
+++ b/src/firewheel/control/model_component_install.py
@@ -149,11 +149,14 @@ class ModelComponentInstall:
 
         Returns:
             bool: :py:data:`True` if the playbook executed successfully, :py:data:`False` otherwise.
+
+        Raises:
+            ValueError: If an invalid ansible.cash_type is provided in the FIREWHEEL config.
         """
         console = Console()
 
         ansible_config = {
-            "ansible_remote_tmp" : config["system"]["default_output_dir"],
+            "ansible_remote_tmp": config["system"]["default_output_dir"],
         }
 
         # Add any remaining configuration options provided in the FIREWHEEL
@@ -170,28 +173,36 @@ class ModelComponentInstall:
             cached_files = []
 
             # Read the playbook file to get the cached_files variables
-            with open(install_path, 'r') as file:
+            with open(install_path, "r") as file:
                 playbook = yaml.safe_load(file)
 
             # Extract variables from the playbook
             variables = {}
             for play in playbook:
-                if 'vars' in play:
-                    variables.update(play['vars'])
+                if "vars" in play:
+                    variables.update(play["vars"])
 
             cached_files = variables.get("cached_files", [])
 
             # Now we need to update the destination path for all the cached files
             # we can prepend the directory of the original install file.
             for file in cached_files:
-                file["destination"] = str(install_path.parent / Path(file["destination"]))
+                file["destination"] = str(
+                    install_path.parent / Path(file["destination"])
+                )
 
             # Call the cache playbook from the ansible_playbooks directory
-            cache_playbook_path = Path(__file__).resolve().parent / Path(f"ansible_playbooks/{cache_type}.yml")
+            cache_playbook_path = Path(__file__).resolve().parent / Path(
+                f"ansible_playbooks/{cache_type}.yml"
+            )
 
             if not cache_playbook_path.exists():
                 # Get a list of all cache types
-                available_types = [file.stem for file in cache_playbook_path.parent.iterdir() if file.is_file()]
+                available_types = [
+                    file.stem
+                    for file in cache_playbook_path.parent.iterdir()
+                    if file.is_file()
+                ]
                 console.print(
                     f"[b red]Failed to find cache_type=[cyan]{cache_type}[/cyan]."
                     f"Available types are: [magenta]{available_types}[/magenta]"
@@ -211,9 +222,9 @@ class ModelComponentInstall:
 
             # Now we need to ensure the config has these properties
             # so that it will properly take in values passed.
-            rc.input_fd=sys.stdin
-            rc.output_fd=sys.stdout
-            rc.error_fd=sys.stderr
+            rc.input_fd = sys.stdin
+            rc.output_fd = sys.stdout
+            rc.error_fd = sys.stderr
             cache_runner = ansible_runner.Runner(config=rc)
 
             # Ensure we are using subprocess mode
--- a/src/firewheel/control/model_component_install.py
+++ b/src/firewheel/control/model_component_install.py
@@ -189,6 +189,15 @@ class ModelComponentInstall:
             # Call the cache playbook from the ansible_playbooks directory
             cache_playbook_path = Path(__file__).resolve().parent / Path(f"ansible_playbooks/{cache_type}.yml")
 
+            if not cache_playbook_path.exists():
+                # Get a list of all cache types
+                available_types = [file.stem for file in cache_playbook_path.parent.iterdir() if file.is_file()]
+                console.print(
+                    f"[b red]Failed to find cache_type=[cyan]{cache_type}[/cyan]."
+                    f"Available types are: [magenta]{available_types}[/magenta]"
+                )
+                raise ValueError("Available `cache_type` are: {available_types}")
+
             # By defining everything here, we can enable prompt's as we will no longer
             # use pexpect by default, but rather use subprocess, enabling stdin
             # See: https://github.com/ansible/ansible-runner/issues/1399
--- a/src/firewheel/control/model_component_install.py
+++ b/src/firewheel/control/model_component_install.py
@@ -10,6 +10,7 @@ from rich.syntax import Syntax
 from rich.console import Console
 
 from firewheel.lib.log import Log
+from firewheel.config import config
 
 
 class InstallPrompt(PromptBase[str]):
@@ -151,21 +152,24 @@ class ModelComponentInstall:
         """
         console = Console()
 
-        # TODO: Pull this information from the FW configuration
         ansible_config = {
-            "ansible_remote_tmp" : "/scratch/firewheel",
-            "cache_type" : "git",
-            #"cache_type" : "s3",
+            "ansible_remote_tmp" : config["system"]["default_output_dir"],
         }
-        # Get the cache_type with a default value of an empty string
+
+        # Add any remaining configuration options provided in the FIREWHEEL
+        # configuration
+        ansible_config.update(config["ansible"])
+
+        # Get the cache_type and default to an empty string
+        # for easy access
         cache_type = ansible_config.get("cache_type", "")
 
-        # Check to see what mode we need to use
-        if cache_type != "online":  # Adjust this condition as needed
-            # Get the cached files
+        # If we set a different cache type, then we should use it
+        # to download the needed cached files.
+        if cache_type != "online":
             cached_files = []
 
-            # Read the playbook file
+            # Read the playbook file to get the cached_files variables
             with open(install_path, 'r') as file:
                 playbook = yaml.safe_load(file)
 
@@ -176,6 +180,11 @@ class ModelComponentInstall:
                     variables.update(play['vars'])
 
             cached_files = variables.get("cached_files", [])
+
+            # Now we need to update the destination path for all the cached files
+            # we can prepend the directory of the original install file.
+            for file in cached_files:
+                file["destination"] = str(install_path.parent / Path(file["destination"]))
 
             # Call the cache playbook from the ansible_playbooks directory
             cache_playbook_path = Path(__file__).resolve().parent / Path(f"ansible_playbooks/{cache_type}.yml")
@@ -202,7 +211,10 @@ class ModelComponentInstall:
             cache_runner.runner_mode = "subprocess"
 
             ret = cache_runner.run()
-            if ret.rc == 0:
+
+            # Invoking the runner this way uses an (unnamed) tuple as a return value
+            # For example: ("success", 0)
+            if ret[1] == 0:
                 console.print(
                     f"[b green]Successfully collected cached files via: [cyan]{cache_type}[/cyan]!"
                 )
--- a/src/firewheel/control/model_component_install.py
+++ b/src/firewheel/control/model_component_install.py
@@ -102,6 +102,10 @@ class ModelComponentInstall:
         if self.is_ansible_playbook(install_path):
             return self.run_ansible_playbook(install_path)
 
+        console.print(
+            "[b yellow]Warning: Use of non-ansible-based INSTALL scripts is deprecated. "
+            "Please contact the model component developer to transition this file."
+        )
         try:
             # We have already checked with the user to ensure that they want to run
             # this script.
--- a/src/firewheel/control/model_component_install.py
+++ b/src/firewheel/control/model_component_install.py
@@ -9,8 +9,8 @@ from rich.prompt import PromptBase
 from rich.syntax import Syntax
 from rich.console import Console
 
-from firewheel.lib.log import Log
 from firewheel.config import config
+from firewheel.lib.log import Log
 
 
 class InstallPrompt(PromptBase[str]):
@@ -173,7 +173,7 @@ class ModelComponentInstall:
             cached_files = []
 
             # Read the playbook file to get the cached_files variables
-            with open(install_path, "r") as file:
+            with open(install_path, "r", encoding="utf-8") as file:
                 playbook = yaml.safe_load(file)
 
             # Extract variables from the playbook
@@ -207,7 +207,7 @@ class ModelComponentInstall:
                     f"[b red]Failed to find cache_type=[cyan]{cache_type}[/cyan]."
                     f"Available types are: [magenta]{available_types}[/magenta]"
                 )
-                raise ValueError("Available `cache_type` are: {available_types}")
+                raise ValueError(f"Available `cache_type` are: {available_types}")
 
             # By defining everything here, we can enable prompt's as we will no longer
             # use pexpect by default, but rather use subprocess, enabling stdin
@@ -241,7 +241,8 @@ class ModelComponentInstall:
                 return True
             else:
                 console.print(
-                    f"[b red]Failed to collect cached files via: [cyan]{cache_type}[/cyan]; Failed with return code {ret.rc}."
+                    f"[b red]Failed to collect cached files via: [cyan]{cache_type}[/cyan]; "
+                    f"Failed with return code {ret.rc}."
                 )
                 return False
 
--- a/src/firewheel/control/model_component_install.py
+++ b/src/firewheel/control/model_component_install.py
@@ -151,7 +151,67 @@ class ModelComponentInstall:
         """
         console = Console()
 
-        ansible_config = None
+        # TODO: Pull this information from the FW configuration
+        ansible_config = {
+            "ansible_remote_tmp" : "/scratch/firewheel",
+            "cache_type" : "git",
+            #"cache_type" : "s3",
+        }
+        # Get the cache_type with a default value of an empty string
+        cache_type = ansible_config.get("cache_type", "")
+
+        # Check to see what mode we need to use
+        if cache_type != "online":  # Adjust this condition as needed
+            # Get the cached files
+            cached_files = []
+
+            # Read the playbook file
+            with open(install_path, 'r') as file:
+                playbook = yaml.safe_load(file)
+
+            # Extract variables from the playbook
+            variables = {}
+            for play in playbook:
+                if 'vars' in play:
+                    variables.update(play['vars'])
+
+            cached_files = variables.get("cached_files", [])
+
+            # Call the cache playbook from the ansible_playbooks directory
+            cache_playbook_path = Path(__file__).resolve().parent / Path(f"ansible_playbooks/{cache_type}.yml")
+
+            # By defining everything here, we can enable prompt's as we will no longer
+            # use pexpect by default, but rather use subprocess, enabling stdin
+            # See: https://github.com/ansible/ansible-runner/issues/1399
+            rc = ansible_runner.RunnerConfig(
+                private_data_dir=str(ansible_config["ansible_remote_tmp"]),
+                playbook=str(cache_playbook_path),
+                extravars={"cached_files": cached_files, **ansible_config},
+            )
+
+            rc.prepare()
+
+            # Now we need to ensure the config has these properties
+            # so that it will properly take in values passed.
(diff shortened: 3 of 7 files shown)


docs/source/model_component/mc_install.rst
```diff
@@ -5,24 +5,128 @@ Model Component INSTALL file
...
+*****************
+Design Principles
+*****************
 
-When a repository is installed via the :ref:`helper_repository_install` Helper, users have the option to can automatically run each MCs INSTALL script using the ``-s`` flag (see :ref:`helper_repository_install` for more details).
+In addition to the requirements above, we recommend that the following principles are adhered to when creating a new INSTALL file.
+
+1. `Idempotence <https://en.wikipedia.org/wiki/Idempotence>`_ - The file should be capable of running multiple times without causing issues. This is a core tenant of Ansible and a strong motivator why Ansible Playbooks are the preferred INSTALL file method.
+2. **Reproducibility** - It is critical that users will download the exact same data that was originally intended by the Model Component creators.
+   If the data/packages differ, than there is a strong possibly that the experimental outcomes will differ and could produce unintended consequences.
```
gregjacobus, line 29: possibly -> possibility

docs/source/model_component/mc_install.rst
```diff
@@ -5,24 +5,128 @@ Model Component INSTALL file
...
+    - name: Download DNS Data files
+      hosts: localhost
+      become: yes
+      vars:
+        install_flag: "{{ playbook_dir }}/.dns.dns_objects.installed"
+        cached_files:
+          - source: "firewheel_repo_dns/dns_objects/bind9_xenial_debs.tgz"
+            destination: "vm_resources/bind9_xenial_debs.tgz"
+
+When designing the tasks, recall that it is possible that the existing ``cached_files`` to become available prior to playbook execution.
```
gregjacobus, line 71: I have no idea what this line is saying

docs/source/model_component/mc_install.rst
```diff
@@ -5,24 +5,128 @@ Model Component INSTALL file
...
+- ``url_cache_path`` - The path to the file's base directory from the server. For example: ``http://example.com/<url_cache_path>/file.txt``.
+- ``use_proxy`` - (optional) If ``false``, it will not use a proxy, even if one is defined in an environment variable on the target hosts. The default is: ``true``.
+- ``validate_certs`` - (optional) If ``false``, SSL certificates will not be validated. The default is: ``true``.
+
+If you are using an username or password token, you can specify it in the URL.
+For example: ``https://user:password@server.com/url/cache/path/file.txt``
+
+Git Cache
+=========
+If users have access to an git server instance, they can specify the ``ansible.cache_type`` key as ``git`` in the :ref:`firewheel_configuration`.
```
gregjacobus, line 109: an -> a

docs/source/model_component/mc_install.rst
```diff
@@ -5,24 +5,128 @@ Model Component INSTALL file
...
+Additional configuration options under the ``ansible`` key are also necessary.
+If these values are not provided, but ``ansible.cache_type`` is ``git``, the user will be prompted for the information.
+
+- ``git_server`` - The URL of the git server.
+- ``git_repo_path`` - The path to the repo from the server. Because this is likely to change for each model component, we recommend not setting this parameter and simply prompting the user for each path.
+- ``git_branch`` - (optional) The branch name, defaults to ``main``.
+
+If you an access token is being used, the user can specify it in the ``git_server`` URL.
```
gregjacobus, line 117: get rid of "you"

src/firewheel/control/model_component_install.py
```diff
@@ -109,6 +123,157 @@ def install_mc(self, name, install_path):
...
+                # For example: ("success", 0)
+                if ret[1] == 0:
+                    console.print(
+                        f"[b green]Successfully collected cached files via: [cyan]{cache_type}[/cyan]!"
+                    )
+                else:
+                    console.print(
+                        f"[b red]Failed to collect cached files via: [cyan]{cache_type}[/cyan]; "
+                        f"Failed with return code {ret[1]}."
+                    )
```
gregjacobus, line 252: I haven't run this yet, but it _feels_ like we're masking the true error when running this way. Do we need to also print `ret[0]`, is that the actual error message?

gregjacobus (changes requested):
Going to test this pretty thoroughly, but here's some comments from a first pass

gregjacobus:
I expect we also want to change the INSTALL file that automatically gets generated with `firewheel mc generate` to be an ansible template? looks like it's still the bash one

src/firewheel/control/model_component_install.py
```diff
@@ -86,13 +89,24 @@ def install_mc(self, name, install_path):
             installed), False otherwise
         """
         console = Console()
+        console.print(f"[b green]Starting to install [cyan]{name}[/cyan]!")
+
+        # Check if the install script is a valid Ansible playbook
+        # if it is, we should execute it with "ansible-runner"
+        if self.is_ansible_playbook(install_path):
```
gregjacobus, line 96: I think it might help if we reverse the logic here. The current problem is, what if you're making an ansible runner yaml file but have a simple syntax error in your yaml? The yaml load fails, so it tries to execute it as an executable. Maybe even more importantly, it tells you absolutely nothing about the small syntax error in your yaml.

What if, instead of checking if it's yaml, we instead check for a shebang? If it has a shebang, run it accordingly. Otherwise, _assume_ that it's trying to be ansible yaml. Especially if we're "deprecating" the shebang install files.

pyproject.toml
```diff
@@ -54,6 +54,7 @@ classifiers = [
     "Topic :: System :: Operating System",
 ]
 dependencies = [
+    "ansible-runner>=2.0.0,<=2.4.1",
```
gregjacobus, line 57: I _think_ regular ansible is also required. I got an error about ansible-playbook not being found when i had only installed ansible-runner

docs/source/model_component/mc_install.rst
```diff
@@ -5,24 +5,128 @@ Model Component INSTALL file
...
+            destination: "vm_resources/bind9_xenial_debs.tgz"
+
+When designing the tasks, recall that it is possible that the existing ``cached_files`` to become available prior to playbook execution.
+
+************
+Cached Files
+************
+
+FIREWHEEL supports collecting pre-computed blobs from various resources to enable offline experiment access.
+To enable retrieving files from a cache, users should set the ``ansible.cache_type`` to ``git`` or ``s3`` depending on if the files are cached in a git repository or in an Amazon S3 data store.
```
gregjacobus, line 78: Still working through the entire architecture of this, so forgive is this is wrong -- but doesn't this force all model components to use the same cache_type? That feels pretty limiting, but maybe there's a way around it that I'm not seeing yet

src/firewheel/control/model_component_install.py
sdelliot: Excellent point! I went the route of showing the entire tuple so that the user gets the message and return code.

docs/source/model_component/mc_install.rst
sdelliot: That is sort of correct and this discussion is exactly why I left this PR in draft mode :). Generally people can call the INSTALL scripts in two different ways. They can execute the model component (via `firewheel experiment`) which will check if the MC needs to be installed. Alternatively, users can run `firewheel repository install --script` which checks for `INSTALL` files in all repositories and runs them.

You are correct that each set of model components that is installed at once must use the same `cache_type`. However, it is feasible modify the `cache_type` between installing another set of components. So it is rather inflexible for a given set of components, but when installing a new repo, can be modified.

The primary (possibly incorrect) assumption is that organizations that wish to cache files will likely use the same methodology for all MC repos. If an organization uses multiple cache types that would likely become a maintenance burden at some point. However, if different groups within an organization were responsible for managing different MC repos, than it seems very feasible that multiple cache types are needed. In this case, the thought was that users would install one set and then install the second set, modifying the `cache_type` in between.

It seems reasonable to address this limitation, but it would likely add some complexity to the system. Without too much prior thought a new CLI argument for `repository install` may be used to override this variable "on the fly". However, that wouldn't address `firewheel experiment`.  Thoughts and discussion are much appreciated!

src/firewheel/control/model_component_install.py
sdelliot: I like this suggestion a lot!

sdelliot: I did a prototype of this now, but I will still need to fix up some of the test cases.

sdelliot:
> I expect we also want to change the INSTALL file that automatically gets generated with `firewheel mc generate` to be an ansible template? looks like it's still the bash one

Yes, this should probably be addressed by this PR. 

- [ ] Update the INSTALL file template

src/firewheel/control/ansible_playbooks/git.yml
```diff
@@ -0,0 +1,41 @@
...
+      private: false
+
+    - name: git_repo_path
+      prompt: "Please enter the path to the git repository or set `ansible.git_repo_path` in the FIREWHEEL configuration."
+      private: false
+
+  tasks:
+    - name: Clone the Git repository
+      ansible.builtin.git:
+        repo: "{{ git_server }}/{{ git_repo_path }}"
```
gregjacobus, line 24: joining git_server and git_repo_path like this means we're requiring cloning with https. SSH cloning will not work I believe

src/firewheel/control/ansible_playbooks/git.yml
```diff
@@ -0,0 +1,41 @@
...
+        dest: "/tmp/fwcache"
+        version: "{{ git_branch | default('main') }}"
```
gregjacobus, line 26: Looks like doing it this way will make it default to whatever the default branch is. So, if the default branch is still master, it will pull that. If git_branch is specified, it will use that
```suggestion
        version: "{{ git_branch | default(omit) }}"
```

docs/source/model_component/mc_install.rst
```diff
@@ -5,24 +5,130 @@ Model Component INSTALL file
...
+For example: ``https://user:password@server.com/url/cache/path/file/path/file.txt``
+
+Git Cache
+=========
+If users have access to a git server instance, they can specify the ``ansible.cache_type`` key as ``git`` in the :ref:`firewheel_configuration`.
+Additional configuration options under the ``ansible`` key are also necessary.
+If these values are not provided, but ``ansible.cache_type`` is ``git``, the user will be prompted for the information.
+
+- ``git_server`` -- The URL of the git server.
+- ``git_repo_path`` -- The path to the repo from the server. Because this is likely to change for each model component, we recommend not setting this parameter and simply prompting the user for each path.
```
gregjacobus, line 116: the `git_repo_path` parameter should really go in the INSTALL file. Relying on a user specifying the git repo is very non-conducive to automated installation

gregjacobus (changes requested):

src/firewheel/control/model_component_install.py
```diff
@@ -109,6 +123,150 @@ def install_mc(self, name, install_path):
...
+                console.print(
+                    f"[b yellow]Cache type [cyan]{cache_type}[/cyan] provided but no cached files "
+                    f"found in [cyan]{install_path}[/cyan]. Continuing."
+                )
+        # Run the Ansible playbooks
+        ret = ansible_runner.run(
+            private_data_dir=str(install_path.parent),
+            playbook=str(install_path),
+            extravars=ansible_config,
+        )
```
gregjacobus, line 257: Is there a reason we run ansible two different times? It feels like the better way is for the INSTALL file to inherit from the cached_type.yml (e.g. git.yml) and run everything together. I think this buys us:

1. There aren't two runs, so the ansible `PLAY RECAP` is all in one place.
2. I think it'll allow us to perform more customization within the INSTALL file. For example, specifying the git_repo_path inside the INSTALL file

src/firewheel/control/model_component_install.py
```diff
@@ -109,6 +123,150 @@ def install_mc(self, name, install_path):
...
+
+            # Read the playbook file to get the cached_files variables
+            with install_path.open("r", encoding="utf-8") as file:
+                playbook = yaml.safe_load(file.read())
+
+            # Extract variables from the playbook
+            variables = {}
+            for play in playbook:
+                if "vars" in play:
+                    variables.update(play["vars"])
```
gregjacobus, line 182: It seems like the intent here is to get variables defined in the INSTALL file and use them for pulling files from the cache. However, this variable is currently only used to pull cache_files.

Actually, it seems like everything except `cached_files` is unused in the INSTALL file's vars. Even the `install_flag` is only using `.{mc_name}.installed` as far as I can tell (and a test with a different install_flag confirmed)

gregjacobus (changes requested):

src/firewheel/control/ansible_playbooks/git.yml
sdelliot: My original tests were with an SSH-based clone. SSH cloning is supported but the [SCP-style URLs](https://git-scm.com/book/ms/v2/Git-on-the-Server-The-Protocols) are not. That is, `git clone ssh://git@github.com/sandialabs/firewheel.git` works just fine. It We could definitely improve the documentation to clarify that though. 

[GitLab mirroring](https://cee-gitlab.sandia.gov/help/user/project/repository/mirror/index) uses the following: 

> NOTE: SCP-style URLs are not supported. 

Or this [GitLab PR ](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/118172) has a more explicit:
> When using the `ssh://` protocol, please use the following format: `ssh://username@example.com/group/project.git`.

I think a combination of those in the documentation would likely suffice. If you think that we should support SCP-style URLs, we could do something like the following.

However, I am not sure what side-effects might happen if we ignore errors for the "HTTPS" clone. We can obviously switch the logic, but would the extra delay be "worth it" for all users?

```diff
diff --git a/src/firewheel/control/ansible_playbooks/git.yml b/src/firewheel/control/ansible_playbooks/git.yml
--- a/src/firewheel/control/ansible_playbooks/git.yml
+++ b/src/firewheel/control/ansible_playbooks/git.yml
@@ -19,7 +19,7 @@
       private: false
 
   tasks:
-    - name: Clone the Git repository
+    - name: Clone the git repository via HTTPS
       ansible.builtin.git:
         repo: "{{ git_server }}/{{ git_repo_path }}"
         dest: "/tmp/fwcache"
@@ -27,6 +27,17 @@
         clone: true
         update: true
       register: git_clone_result
+      ignore_errors: yes
+
+    - name: Clone the git repository via SSH (if needed)
+      ansible.builtin.git:
+        repo: "git@{{ git_server }}:{{ git_repo_path }}"
+        dest: "/tmp/fwcache"
+        version: "{{ git_branch | default('main') }}"
+        clone: true
+        update: true
+      when: git_clone_result is failed
+      register: git_clone_result
 
     - name: Move cached files to the desired location
       ansible.builtin.copy:
```

docs/source/model_component/mc_install.rst
sdelliot: Maybe this is partly a variable name issue. ``git_repo_path`` is not the exact git repo, but rather the group name + the repo name. For this repo here is what it would look like:

- `git_server=https://github.com`
- `git_repo_path=sandialabs/firewheel.git`

I suspect that maybe this should be in three different variables. It doesn't make sense for the `INSTALL` files to specify the group name, but maybe forcing the same "repo name" makes sense?


- `git_server=https://github.com`
- `git_group_path=sandialabs`
- `git_repo_name=firewheel.git` - Defined in INSTALL.

gregjacobus: That's definitely better, but I still don't love that you can only have binaries stored in one git group. Maybe I'm being too picky, so feel free to push back if you think so.

This may get ugly somewhat ugly, but what if we made some of these arguments arrays? the repo name could be in the INSTALL file like you suggest, but then there could be something like this in the firewheel config:

```yaml
ansible:
    git_servers:
        - https://github.com/sandialabs
        - https://github.com/otherrepo
        - https://gitlab.com/sandialabs
        - https://gitlab.com/otherrepo
```

Then, the system could look in each of these spots to look for the repository specified by `git_repo_name`. The biggest downside I see is that we're now checking a lot of locations, which might make it slow. But, if you only specify one repo, it's actually no slower than it would be in the current method. Thoughts?

gregjacobus:
`firewheel repository install --help` still references an INSTALL _file_

src/firewheel/control/utils/templates/tasks.yml.j2
```diff
@@ -0,0 +1,24 @@
...
+#
+# FIREWHEEL will automatically provide the following variables to Ansible playbooks:
+# - ``mc_name`` -- The name of the Model Component.
+# - ``mc_dir`` -- The full path to the model component directory.
+#
+###################################################################################################
+# Example tasks
+#- name: Create directory for downloading HTOP
+#  ansible.builtin.file:
+#    path: "htop-1_0_2_debs"
```
gregjacobus, line 17:
```suggestion
#    path: "{{mc_dir}}/htop-1_0_2_debs"
```

src/firewheel/control/utils/templates/tasks.yml.j2
```diff
@@ -0,0 +1,24 @@
...
+#    state: directory
+#
+#- name: Download htop Package
+#  ansible.builtin.get_url:
+#    url: "http://archive.ubuntu.com/ubuntu/pool/universe/h/htop/htop_1.0.2-3_amd64.deb"
+#    dest: "htop-1_0_2_debs/htop_1.0.2-3_amd64.deb"
```
gregjacobus, line 23: Looks like absolute paths are needed

```suggestion
#    dest: "{{mc_dir}/htop-1_0_2_debs/htop_1.0.2-3_amd64.deb"
```

src/firewheel/control/utils/templates/vars.yml.j2
```diff
@@ -0,0 +1,31 @@
...
+# Without being as specific as possible, experimental results will **NOT** be repeatable.
+# We strongly recommend that any changes to software versions are accompanied by a warning and a
+# new model component version.
+###################################################################################################
+# If there are files that can be cached, the final file name should
+# be added to this list. Then offline users may use a storage mechanism
+# for quick retrieval, rather than downloading from the Internet each time.
+cached_files:
+#   - source: "<package name>/<path to MC>/file.xz"
+#     destination: "vm_resources/example.tgz"
```
gregjacobus, line 29:
```suggestion
#     destination: "{{mc_dir}}/vm_resources/example.tgz"
```

gregjacobus (changes requested):

docs/source/model_component/mc_install.rst
```diff
@@ -5,34 +5,482 @@ Model Component INSTALL file
...
+
+.. _mc_install_cache:
+
+***********
+Cache Types
+***********
+
+Collecting and retrieving files from a cache is automatically supported in Ansible playbooks without MC designer intervention.
+Currently, FIREWHEEL supports caching files in a file server, git repository, or in an Amazon S3 data store.
+If the user sets the necessary settings in the :ref:`firewheel_configuration` for the described types below, than FIREWHEEL will automatically check those locations for the cached file.
```
gregjacobus, line 169:
```suggestion
If the user sets the necessary settings in the :ref:`firewheel_configuration` for the described types below, then FIREWHEEL will automatically check those locations for the cached file.
```

docs/source/model_component/mc_install.rst
```diff
@@ -5,34 +5,482 @@ Model Component INSTALL file
...
+If no ``cached_files`` are needed, then it can be omitted from ``INSTALL/vars.yml``.
+
+Continuing the example from above, the ``vars.yml`` would look like:
+
+.. code-block:: yaml
+  :caption: This is an example ``vars.yml`` file that ensures the final MC state.
+
+  cached_files:
+    - source: "firewheel_repo_linux/ubuntu/ubuntu/htop-1_0_2_debs.tgz"
+      destination: "{{ mc_dir }}/vm_resources/debs/htop-1_0_2_debs.tgz"
```
gregjacobus, line 115: This example is somewhat confusing. It's supposedly continuing the example from above, but the example above is pulling from a URL. This one seems to be pulling from a git repo. 

What is the cached_files object supposed to look like when downloading from a URL? Specifically, what is the source supposed to be?

Also, adding some of this in feels a bit like duplicating the results from tasks.yml. Is that not true for some reason?

gregjacobus (changes requested):

src/firewheel/control/ansible_playbooks/roles/firewheel/tasks/check_and_end.yml
```diff
@@ -0,0 +1,44 @@
...
+# Create installation flag if all cached files exist
+- name: Create installation flag if all cached files exist
+  ansible.builtin.file:
+    path: "{{ install_flag }}"
+    state: touch
+    mode: "660"
+  when: all_files_valid
+
+# Exit successfully if there are no issues with cached files
+- name: Exit successfully if there are issues with cached files
```
gregjacobus, line 42:
```suggestion
- name: Exit successfully if there are no issues with cached files
```

docs/source/model_component/mc_install.rst
sdelliot: I think the language of `cached_files` is still a bit nebulous. Generally, the `cached_files` source has be the same regardless of the caching method used. That's because the MC designer sets this source of the file. Therefore, the end user has to put the file in that location for their given cache. That's why standardizing the path of the files is critical. All files should follow the convention: `<package name>/<path to MC>/file`. Then the end-user would set up the cache like:

*git* = `my_cache_repo1/<package name>/<path to MC>/file`

*url* = `http://localhost:8000/<package name>/<path to MC>/file`

*S3* = `https://aws.amazon.com/mybucket/<package name>/<path to MC>/file`

I added the following to help clarify how this works.

```diff
-Continuing the example from above, the ``vars.yml`` would look like:
+Continuing the example from above, the end result of ``tasks.yml`` is the creation of the file ``{{ mc_dir }}/vm_resources/debs/htop-1_0_2_debs.tgz``.
+Rather than run the tasks (which assume online access), users could provide this file from a cache.
+The ``vars.yml`` file would look like:
```

I think this is one of the harder and less-well-defined concepts in this entire feature, so any thoughts on how to clarify this are greatly appreciated!

sdelliot:
> `firewheel repository install --help` still references an INSTALL _file_

Fixed in [01a4451](https://github.com/sandialabs/firewheel/pull/113/commits/01a4451088b721635a52f741e9cb9bae7638e8b4)

docs/source/model_component/mc_install.rst
sdelliot: Well, specifically, it is pulling down htop, putting it into a folder, and compressing that folder for the example experiment to use. In that case, the "cached file" would be the compressed folder. It definitely could be the debian package, if that is used directly by the MC. But in either case, the point of the cache is a check that all files that need to exist, do, and for any offline people, they can replicate that. That is part of the odd thing about cached_files, is that they sort of only exist for the purposes of offline use, (especially their "source") but are almost unnecessary for users.

As I think about this, I have some thoughts on making this a bit more user-friendly. What if the `cache_files` don't have a `source` and that the `source` is provided by FIREWHEEL. That will help with consistency of the location. Then we can have a "help" page for setting up a cache. Then to enforce that these cache files must be defined, we can enforce the convention that the script will only pass if all files in the cache are present (and have good checksums if provided). Then MC designers will have to add them but won't have any impact on how the cache is used. That may clear up the confusion around it. We could rename if necessary. Rather than "cache" it could be `required_files` or something like that.

gregjacobus: Not sure I 100% see your vision for FIREWHEEL providing the `source`, but I think I'm close enough to say I support it. And yes, I think `required_files` is much clearer in this case. That feels more intuitive for my understanding of it

docs/source/model_component/mc_install.rst
```diff
@@ -5,34 +5,519 @@ Model Component INSTALL file
 ############################
 
 Some Model Components may require additional Python packages to be installed within FIREWHEEL's virtual environment or for data to be downloaded.
-In this case, the Model Component can have an ``INSTALL`` file, which can be any executable script (as denoted by a `shebang <https://en.wikipedia.org/wiki/Shebang_(Unix)>`_ line).
+In this case, the Model Component can have an ``INSTALL`` directory, which contains a valid `Ansible Playbook <https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_intro.html>`_ (recommend method).
```
gregjacobus, line 8:
```suggestion
In this case, the Model Component can have an ``INSTALL`` directory, which contains a valid `Ansible Playbook <https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_intro.html>`_ (recommended method).
```

docs/source/model_component/mc_install.rst
```diff
@@ -5,34 +5,519 @@ Model Component INSTALL file
...
+Alternatively, ``INSTALL`` can be an executable script (as denoted by a `shebang <https://en.wikipedia.org/wiki/Shebang_(Unix)>`_ line), though this is not recommend and support will be removed in a future releases.
```
gregjacobus, line 9:
```suggestion
Alternatively, ``INSTALL`` can be an executable script (as denoted by a `shebang <https://en.wikipedia.org/wiki/Shebang_(Unix)>`_ line), though this is not recommended and support will be removed in a future releases.
```

docs/source/model_component/mc_install.rst
```diff
@@ -5,34 +5,519 @@ Model Component INSTALL file
...
+When users use the :ref:`helper_mc_generate` Helper, a new INSTALL directory is created with sample ``tasks.yml`` and ``vars.yml`` automatically included.
 
-When creating an INSTALL file, it is critical that users will download the exact same data that was originally intended by the Model Component creators.
-If the data/packages differ, than there is a strong possibly that the experimental outcomes will differ and could produce unintended consequences.
-Therefore, we strongly recommend that MC creators link to exact versions of software to download, rather than an automatically updating link.
-For example, if the MC was supposed to install a GitLab runner:
 
-.. code-block:: bash
+When a repository is installed via the :ref:`helper_repository_install` Helper, users have the option to can automatically run each MCs INSTALL script using the ``-s`` flag (see :ref:`helper_repository_install` for more details).
```
gregjacobus, line 13:
```suggestion
When a repository is installed via the :ref:`helper_repository_install` Helper, users have the option to automatically run each MC's INSTALL script using the ``-s`` flag (see :ref:`helper_repository_install` for more details).
```

gregjacobus (changes requested):
just a few spelling things for now. Still getting through some more architecture review, but so far, looking really good

docs/source/model_component/mc_install.rst
```diff
@@ -5,34 +5,519 @@ Model Component INSTALL file
...
+    :type: string
+    :required: false
+
+
+.. confval:: checksum
+
+    The hash of the file.
+
+    :type: string
+    :required: false
```
gregjacobus, line 154: So one issue we should probably consider is that tar archives do not get consistently hashed. So, specifying a checksum in `required_files` will work if a cached file is pulled, but not if it gets recreated by pulling the files from the regular internet. 

It looks like some fancy arguments to tar can make it consistently hashed, such as the following:

```bash
tar --mtime='2025-01-01' --sort=name --no-xattrs -cf - file1 file2 | gzip --no-name > test1.tgz
```

If the above is what we want to be the "recommended" approach, despite it being ugly, we should probably have an example of doing this with ansible. In the old install files, we made sure to check hashes of the un-compressed files before compressing, but this case is slightly different. Definitely open to better alternatives though

docs/source/model_component/mc_install.rst
```diff
@@ -5,34 +5,519 @@ Model Component INSTALL file
...
+          - path: "firewheel/mc_repo1"
+          - path: "firewheel/mc_repo2"
+            branch: "develop"
+      - server_url: "ssh://git@gitlab.com"
+        repositories:
+          - path: "emulytics/firewheel/mc_repo3"
+            branch: "feature-branch"
+      - server_url: "https://user:ACCESS-TOKEN@github.com/"
+        repositories:
+          - path: "firewheel/mc_repo4"
```
gregjacobus, line 194: Okay I think I found a place we'll really have to optimize. It looks like EACH git repository gets cloned for EACH model component install that happens. Plus, each time it's cloned, it gets subsequently cleaned up. That means that if we're attempting to download binaries for 5 model components (pretty normal), one of which is linux, we have to clone the linux binary repo 5 times. This is going to be painfully slow and isn't necessary. So, a couple options:

1. Check to see if the required_files is satisfied after cloning each repository. That helps somewhat, but repos are still needlessly cloned frequently. We could put the biggest repos at the end so this doesn't happen as much, but still not great
2. Don't clean up cached repositories after they're cloned and don't re-clone if it's already cached. But does this create any implications for old versions of repositories stored in the cache? 
3. Hear me out... what if we don't do an LFS pull automatically? We do a non-lfs pull, check to see if the files we need are there, and then manually pull them if they exist. 

I'm open to other options, but I think as it is right now is going to be really difficult

src/firewheel/control/ansible_playbooks/roles/firewheel/tasks/git_clone_loop.yml
```diff
@@ -0,0 +1,45 @@
...
+###################################################################################################
+
+- name: Clone the repository
+  ansible.builtin.git:
+    repo: "{{ firewheel_server_url }}/{{ firewheel_repo_path }}"
+    dest: "{{ ansible_remote_tmp }}/git"
+    clone: true
+    version: "{{ firewheel_repo_branch }}"
+    update: true
+  register: git_clone_result
```
sdelliot, line 31: Per the suggestion from @gregjacobus we will try to not pull LFS objects unless they are necessary for a specific required file.

docs/source/model_component/mc_install.rst
sdelliot: This is a great suggestion and I went with option 3. I ran a test with a non-optimal git cache configuration (where large repos without any of the required files were pulled first) and it took 477.77s while the new approach only took 187.11s, a significant time savings (290.66 seconds)! The new approach and updated documentation have now been pushed.

docs/source/model_component/mc_install.rst
sdelliot: I added some documentation to note these issues.

gregjacobus:
Not sure if this should be considered a "bug" or not, but mentioning here at least for discussion. I have the following `tasks.yml` file:

```yml
- name: Create VM resources directory
  ansible.builtin.file:
    path: "{{ mc_dir }}/vm_resources"
    state: directory
```

And the following `vars.yml` file:

```yml
required_files:
   - destination: "{{ mc_dir }}/vm_resources/bind9_xenial_debs.tgz"
     source: "dns.dns_objects/vm_resources/bind9_xenial_debs.tgz"
```

I have a git cache setup to pull the debs. However, this fails if the `vm_resources` directory doesn't exist, since the tasks.yml (which creates it) doesn't run until after the cache is checked.

So really the issue is that, from the same starting state, pulling from the internet and pulling from a cache don't necessarily both work

gregjacobus:
More of a general comment, and maybe not something we need to hold up this PR for -- but I found this somewhat hard to debug. Some of the success vs. failure of tasks seems slightly unintuitive to me. For example, there's a task that "Checks for required files in the repository". This task "succeeds" even if the file is not found. I think there's a couple others like this that I'm happy to help track down if it's something we think we should address

sdelliot:
> Not sure if this should be considered a "bug" or not, but mentioning here at least for discussion. I have the following `tasks.yml` file:
> 
> ```yaml
> - name: Create VM resources directory
>   ansible.builtin.file:
>     path: "{{ mc_dir }}/vm_resources"
>     state: directory
> ```
> 
> And the following `vars.yml` file:
> 
> ```yaml
> required_files:
>    - destination: "{{ mc_dir }}/vm_resources/bind9_xenial_debs.tgz"
>      source: "dns.dns_objects/vm_resources/bind9_xenial_debs.tgz"
> ```
> 
> I have a git cache setup to pull the debs. However, this fails if the `vm_resources` directory doesn't exist, since the tasks.yml (which creates it) doesn't run until after the cache is checked.
> 
> So really the issue is that, from the same starting state, pulling from the internet and pulling from a cache don't necessarily both work

I think that this should be considered a bug and shouldn't be too challenging to address.

PR #114, opened 2025-04-18
fix: Clearly delineate between editable/non-editable MC installs

mitchnegus:
This captures a handful of the higher priority elements introduced by #91, while focusing on backwards compatibility. 

The primary modification here is to make more clear the distinction between cloned/uncloned editable/non-editable installs and have the docs adequately reflect the actions in the `install.sh` script. Previously, the documentation stated that the installer would clone our default model components, which was true before we transitioned to the package model. As we've made that transition, some of the implied functionality has been lost—right now, running the install script with no arguments will clone MCs but not install them... so still technically accurate, but somewhat misleading because we do not mention that any further actions to be taken. In a similar way, calling the script with `--no-clone` will also not clone MCs, but will also not install them even though one could argue that a familiar user (e.g., me) might expect `--no-clone` to not _clone_ the MC repos, but still install them using pip. There are a few other similar logic cases, but I won't discuss them at length. Instead, I'll summarize the behavior of the script as would be defined by this MR:

|  CLI Option          | (clone)  | `--no-clone` |
|------------------|---------|---------------|
| (production)         | pip install model components from cloned versions | pip install model components from PyPI  |
| `--development` | pip install model components from cloned versions in editable mode | pip install model components from PyPI (same as above) |

In all cases, when using this script, FIREWHEEL will be installed from this local version not PyPI. I don't think that's actually the case right now (looking at the line in the `install_firewheel` function), but this seems like it would be the expected behavior of running this script. Presumably, whoever's running the install script in the repo is doing it because they want to install that version of FIREWHEEL, not just have it turn around and download whatever the latest version is on PyPI.

### DIFF
--- a/docs/source/install/install.rst
+++ b/docs/source/install/install.rst
@@ -192,7 +192,7 @@ Once the configuration options have been set, run ``install.sh``:
     chmod +x install.sh
     ./install.sh
 
-This script will create necessary directories, installs and configures FIREWHEEL, and clone our default model component repositories.
+This script will create necessary directories, install and configure FIREWHEEL, and install our default model component repositories.
 The script will optionally install additional FIREWHEEL development dependencies by using either the ``-d`` or ``--development`` flag::
 
     ./install.sh -d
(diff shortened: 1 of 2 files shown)


docs/source/install/install.rst
```diff
@@ -207,10 +207,23 @@ Once the entire cluster has been provisioned, we recommend running the :ref:`com
...
+.. code-block:: bash
+
+   pip install firewheel_repo_base
+
+A selection of these MCs are installed by the ``install.sh`` script.
+
+The second way to install a Model Component is using FIREWHEEL's helpers.
+This is a good solution for Model Component repositories that are either built or cloned to a local directory.
+We recommend putting all these Model Component repositories in ``/opt/firewheel/model_components``.
+Then you can use the *install* helper to add this location as a known FIREWHEEL Model Component repository::
```
sdelliot, line 226: This can be linked to the correct helper with :ref:`helper_repository_install`

sdelliot (changes requested):

docs/source/install/install.rst
mitchnegus, line 226:
```suggestion
Then you can use the :ref:`helper_repository_install` to add this location as a known FIREWHEEL Model Component repository::
```

gregjacobus (approved):
Haven't personally tested it, but looks good as far as I can tell. Let me know if you need a more thorough round of testing

sdelliot (approved):
This all looks good to me and should make the installer more accurate and consistent.

Issue #119, opened 2025-04-24
Move GitLab CI to directory


mitchnegus:
We have all the GitLab CI descriptions in the `.gitlab-ci.yml` file which is the default location for GitLab CI to find things. If we choose to include that file in other repositories (like MCs) then we are forced into an all-or-nothing situation where every job in the CI file is inherited.

I'm proposing creating a `.gitlab` directory that is analagous to the `.github` directory where we store GitLab CI files (for example, a `.gitlab/setup.yml` file) that only includes the hidden jobs at the top. Then these can be included in other CI files without pulling in _all_ the unit tests, functional tests, linting, deployment, etc. stuff that we have. We've shortened that file considerably since moving primarily to GitHub CI, but this also makes the file a bit more manageable.

Alternatively, we could just discourage use of the `include` keyword in the CI functional tests, and instead prefer that jobs are written to reference other files explicitly.

PR #120, opened 2025-04-25
ci: Use additional files and reference tags for CI setup

mitchnegus:
This moves the anchors used for setup into a separate file `.gitlab/ci/setup.yml` (and converts them to references). This allows other GitLab repositories to include that YAML file directly instead of including the main `.gitlab-ci.yml` file and then needing to override or otherwise negate jobs that should not be run.

I also changed a couple of the job names and removed the `virtualenv` dependency for Docker-based jobs in order to reuse a bunch of the references in both the Docker-based jobs (docs, pages, build, deploy) and the ones running in a native runner (performance). I didn't see an obvious reason that we needed `virtualenv` beyond Python's built-in `venv`, and it didn't play nicely with the native runner. If it is a requirement, I'll find a workaround.


mitchnegus:
Closes #119.

gregjacobus (approved):
Looks good to me. This is a low-consequence change if anything is broken anyway.

PR #125, opened 2025-05-05
fix: Fixing namespaces

gregjacobus:
This fixes FIREWHEEL working in a cluster. There are a few places where the minimega namespace was hard-coded.

I cannot figure out why this change is necessary now, but everything seemed to work in the past. Doesn't seem like the minimega API changed at all. Not sure it's necessarily important, but would love to know if anybody has comments.

TODO: 

- [x] Check if minimega API version matters and fix dependency accordingly

### DIFF
--- a/src/firewheel/lib/minimega/api.py
+++ b/src/firewheel/lib/minimega/api.py
@@ -49,11 +49,17 @@ class minimegaAPI:  # noqa: N801
         self.mm_base = mm_base
         self.mm_socket = os.path.join(self.mm_base, "minimega")
 
+        try:
+            namespace = config["minimega"]["namespace"]
+        except KeyError as exp:
+            self.log.warn("minimega namespace not set, using default")
+            namespace = None
+
         if not os.path.exists(self.mm_socket):
             self.log.error("minimega socket does not exist at: %s", self.mm_socket)
             raise RuntimeError(f"minimega socket does not exist at: {self.mm_socket}")
         try:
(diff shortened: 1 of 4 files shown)


gregjacobus:
I feel pretty good about this, but would like another set of eyes before merging. Also, I have no idea why all these linting jobs are failing...

src/firewheel/lib/minimega/api.py
```diff
@@ -49,11 +49,17 @@ def __init__(self, mm_base=None, timeout=120, skip_retry=False):
         self.mm_base = mm_base
         self.mm_socket = os.path.join(self.mm_base, "minimega")
 
+        try:
+            namespace = config["minimega"]["namespace"]
+        except KeyError:
+            self.log.warning("minimega namespace not set, using default")
+            namespace = None
+
```
mitchnegus, line 57: To avoid misleading errors in the case where the `KeyError` is raised because `minimega` is somehow missing from the `config` option instead of `namespace`, could we substitute the following equivalent block?

```suggestion
        if (namespace := config["minimega"].get("namespace")) is None:
            self.log.warning("minimega namespace not set, using default")
```

mitchnegus (changes requested):

gregjacobus:
Pending any other comments, this should be good to go

sdelliot:
I tested this out and it works for me and resolves the issues.

PR #127, opened 2025-05-15
chore: Tightening dependencies to address python 3.8 support

sdelliot:
[Setuptools 77 ](https://github.com/pypa/setuptools/blob/12ca0186ba7d9bf387d65400bb05205d0bcf9e56/NEWS.rst#v7700) added license expression ([PEP 639](https://peps.python.org/pep-0639/#add-license-expression-field)) support. However, this breaks the current method of how setuptools checks for the license (see https://github.com/pypa/setuptools/issues/4903). While upgrading is ideal, given our continued support for Python 3.8, we must instead lock in the version of setuptools. Additionally, this PR locks in several other packages to address similar deprecation and Python 3.8 issues. 

Lastly, this PR addresses some circular dependency challenges which occurred when installing firewheel[dev]. These changes should fix (and stablize) our CI pipeline as well.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -255,13 +255,13 @@ function install_firewheel_development() {
 
     # Install the development version.
     if [[ $clone -eq 0 ]]; then
-        ${PYTHON_BIN} -m pip install ${PIP_ARGS} -e ${FIREWHEEL_ROOT_DIR}/[dev]
+        ${PYTHON_BIN} -m pip install ${PIP_ARGS} -e ${FIREWHEEL_ROOT_DIR}/[dev,mcs]
     else
-        # In this case, we do not use the "dev" optional dependencies as
+        # In this case, we do not use the "mcs" optional dependencies as
         # the user is using the source code version of these model components, rather
         # than the Python package installed repositories.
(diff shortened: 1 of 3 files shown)


mitchnegus:
@sdelliot, I looked into this and added to your comment on #91. I'm still in favor of getting rid of `[mcs]` from `[dev]`, and I'd personally prefer sooner rather than later. 

Removing that bias though, per that other thread I'm not sure it's necessary to include in this fix? Would that knowledge change things in this MR?

mitchnegus (changes requested):
Current set of changes look good. This will require that `[dev]` in the [install script](https://github.com/sandialabs/firewheel/blob/db49ab488655c58eb6a8f5204e214a3bf572d58e/install.sh#L258) be converted to `[dev,mcs]` (so still functionally equivalent). 

It may also make sense to change the [reference](https://github.com/sandialabs/firewheel/blob/db49ab488655c58eb6a8f5204e214a3bf572d58e/install.sh#L264) to `firewheel[docs,format]` a few lines below that to now be `[dev]` (since this PR will change `[dev]` to be equal to `[format,docs]`) and update/delete the comment accordingly.

mitchnegus (approved):

PR #128, opened 2025-05-19
feat(cli): Adding a new firewheel config edit command

sdelliot:
This PR adds a new command which enables users to easily edit the FIREWHEEL configuration. The `firewheel config edit` command will automatically use the `VISUAL` or `EDITOR` environment variables. Alternatively, users can pass in their own editor of choice (or override the environment variables). This PR also updates the associated documentation.

### DIFF
--- a/docs/source/cli/commands.rst
+++ b/docs/source/cli/commands.rst
@@ -35,6 +35,22 @@ Users can interact with the :ref:`command_config` command (i.e. ``firewheel conf
 series of sub-commands which enable easily getting/setting various configuration options.
 
 
+.. _command_config_edit:
+
+config edit
+^^^^^^^^^^^
+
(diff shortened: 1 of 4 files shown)


mitchnegus:
The code looks good and works for me when testing. Do we want to add unit tests for these? (Just by mocking the subprocess call to ensure that it's called with the proper command?)

sdelliot:
> The code looks good and works for me when testing. Do we want to add unit tests for these? (Just by mocking the subprocess call to ensure that it's called with the proper command?)

For sure, updated in [a340a05](https://github.com/sandialabs/firewheel/pull/128/commits/a340a05ea2f41b584b3f31960df92a25fb0cb04f).

mitchnegus (approved):

PR #130, opened 2025-05-22
chore: Remove duplicate pip installation instruction

mitchnegus:
I didn't notice this on #127, but pre-commit and tox are installed by the `dev` extra. Because of that, it is redundant for us to install them before running our local flavor of `pip install firewheel[dev]` in the install script.


sdelliot (approved):
LGTM

PR #134, opened 2025-05-30
feat: Adding new docker build paramameters

sdelliot:
This PR adds new build parameters that make it easy to modify the creation of a FIREWHEEL docker container and corresponding documentation to help override the default configuration options. Additionally, this moves the docker container to use port 2222 by default as this will prevent conflicts with host systems. This convention is used by other virtualization systems, such as Vagrant.

### DIFF
--- a/docker/fsroot/usr/local/bin/entry
+++ b/docker/fsroot/usr/local/bin/entry
@@ -32,7 +32,7 @@ export HOME=/root
 
 HN="$(hostname)"
 
-FILEPATH="/tmp/minimega/files"
+FILEPATH="$(firewheel config get minimega.files_dir)"
 mkdir -p $FILEPATH
 
 grn "Starting minimega...       "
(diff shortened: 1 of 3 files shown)


docker/fsroot/usr/local/bin/entry
```diff
@@ -25,14 +25,15 @@ else
     cat << EOF > ~/.ssh/config
 Host *
     StrictHostKeyChecking no
+    Port 2222
```
gregjacobus, line 28: I'm not _necessarily_ opposed to this, but it does make a dockerized cluster slightly harder. Also, not sure what the intent is for git cloning inside the container, but I think (?) this would mess up cloning via ssh?

Thoughts on this instead?

```suggestion
Host $(hostname)
    StrictHostKeyChecking no
    Port 2222
```

sdelliot: I think narrowing it to just the host makes a lot of sense. For a dockerized cluster, likely this will need to be extended to the entire set of compute nodes. I fixed this in [32450bb](https://github.com/sandialabs/firewheel/pull/134/commits/32450bbaf11b320a60d369fe9ff114c151a23e53).

gregjacobus (approved):

PR #157, opened 2025-09-03
fix: Expand docker variables and fix discovery path

sdelliot:
This PR makes the following adjustments to the Docker container:
* Corrects the path to Discovery
* Includes an additional minimega environment variable (`MM_LOGFILE`)
* Removes hardcoded paths with the respective environment variable
* Creates expected paths for log files
* Updates the `firewheel config`  with any minimega environment variables

### DIFF
--- a/docker/firewheel.dockerfile
+++ b/docker/firewheel.dockerfile
@@ -15,6 +15,7 @@ ARG MM_DEGREE=1
 ARG MM_CONTEXT=firewheel
 ARG MM_FORCE=true
 ARG MM_LOGLEVEL=debug
+ARG MM_LOGFILE=/var/log/minimega.log
 
 ## FIREWHEEL Arguments
 ARG GRPC_HOSTNAME=localhost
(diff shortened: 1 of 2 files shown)


docker/firewheel.dockerfile
```diff
@@ -73,12 +74,15 @@ RUN bash -c "python3.10 -m venv /fwpy \
 
 # Configure Firewheel
 RUN bash -c "source /fwpy/bin/activate  && \
-    mkdir -p /var/log/minimega && \
+    mkdir -p ${LOGGING_ROOT_DIR} && \
+    mkdir -p ${MM_FILEPATH} && \
+    mkdir -p ${OUTPUT_DIR}/vm_resource_logs && \
+    mkdir -p ${OUTPUT_DIR}/transfers && \
     mkdir -p ${OUTPUT_DIR} && \
```
mitchnegus, line 81: Minor suggestion to move creating `OUTPUT_DIR` above the others.

```suggestion
    mkdir -p ${OUTPUT_DIR} && \
    mkdir -p ${OUTPUT_DIR}/vm_resource_logs && \
    mkdir -p ${OUTPUT_DIR}/transfers && \
```
Arguably it could just be removed, because the directory must exist already for `vm_resource_logs` and `transfers` to be created there, and the `-p` option in those commands should create it if it doesn't exist. But I think it makes sense to leave it as an obvious step by itself (i.e., always make sure that the output directory is created).

mitchnegus: Wait, sorry, did you have a reason for leaving the `mkdir -p ${OUTPUT_DIR}` step after the other two though? It seems like if that's a step by itself it should be first unless I'm missing something

mitchnegus (approved):

PR #162, opened 2025-09-09
ci: Updating the GitLab CI for ansible usage

sdelliot:
This is preparation for the next FW release and will not work until new releases of model components have all been tagged. This has already been tested for syntax errors and will _likely_ work as expected.


gregjacobus (approved):
I really hate the `sed`, but I don't think there's a better alternative at this point. We can update later

PR #165, opened 2025-09-10
chore: Remove code for building wheel in favor of CI deployments

mitchnegus:
This removes the call to build a FIREWHEEL package source distribution from the install script. This should no longer be necessary now that FIREWHEEL is either installed directly from PyPI or via an editable install.

For applications where building a source distribution is necessary (e.g., publishing to PyPI), that process should be handled elsewhere—such as in CI job or in a local build workflow.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -195,30 +195,6 @@ function check_deps() {
 }
 
 #######################################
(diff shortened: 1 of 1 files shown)


sdelliot (approved):
LGTM!

PR #166, opened 2025-09-10
feat: Setting configuration with JSON string

sdelliot:
This PR adds the ability for users to take in a properly formatted JSON string and set the FIREWHEEL configuration. 

This JSON string will set/replace a _subset_ of the configuration and any keys or sub-keys (or sub-sub-keys) not present **will** not be impacted or overwritten. 
The string _must_ include the top-level config key as well as any sub-keys needing modifications.

For example, to change the value for the config key ``logging.level`` from DEBUG to INFO, here is how that would not look:

**Before Config**
```yaml
logging:
    cli_log: cli.log
    discovery_log: discovery.log
    firewheel_log: firewheel.log
    level: DEBUG
    minimega_log: minimega.log
    root_dir: /scratch/firewheel
    vmr_log_dir: vm_resource_logs
```

```bash
firewheel config set -j '{"logging":{"level":"INFO"}}'
```

**After Config**
```yaml
logging:
    cli_log: cli.log
    discovery_log: discovery.log
    firewheel_log: firewheel.log
    level: INFO
    minimega_log: minimega.log
    root_dir: /scratch/firewheel
    vmr_log_dir: vm_resource_logs
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ dependencies = [
     "colorama<=0.4.6",
     "coverage<=7.10.7",
     "grpcio>=1.63.0,<=1.76.0",
-    "grpcio-tools>=1.63.0,<=1.75.1",
+    "grpcio-tools>=1.63.0,<=1.76.0",
     "importlib_metadata>=7.0.0,<=8.7.0",
     "Jinja2>=3.1.2,<=3.1.6",
     "netaddr<=1.3.0,>=0.7.0",


sdelliot:
> > Not exactly sure why, but this doesn't allow spaces in the json string. For example, the following command fails saying it's invalid json, when it actually is valid:
> > ```
> > firewheel config set -j '{"ansible":{"git_servers": []}}'
> > ```
> 
> Looks like [the input arguments get split as strings on whitespace rather than as shell inputs](https://github.com/sandialabs/firewheel/blob/7502f13af85357ab51c7262351fc0a71049c7285/src/firewheel/cli/configure_firewheel.py#L46), so quoted inputs are ignored and your string gets parsed as just `'{"ansible":{"git_servers":`.

Following up on this, this is resolved as of #207 and I have verified that spaces now work.

docs/source/cli/helper_docs.rst
```diff
@@ -725,16 +725,18 @@ The repository should be an existing directory on the filesystem.
 The path may be specified absolute or relative.
 If the path does not exist, an error message is printed.
 
-Some Model Components may provide an additional install script called ``INSTALL`` which can be executed to perform other setup steps (e.g. installing an extra python package or downloading an external VM resource).
-INSTALL scripts can be can be any executable file type as defined by a `shebang <https://en.wikipedia.org/wiki/Shebang_(Unix)>`_ line.
+Some Model Components may provide an additional installation details which can be executed to perform other setup steps (e.g. installing an extra python package or downloading an external VM resource).
```
mitchnegus:
```suggestion
Some Model Components may provide additional installation details which can be executed to perform other setup steps (e.g., installing an extra python package or downloading an external VM resource).
```

docs/source/cli/helper_docs.rst
```diff
@@ -725,16 +725,18 @@ The repository should be an existing directory on the filesystem.
...
+This takes the form of either a ``INSTALL`` directory with a ``vars.yml`` and a ``tasks.yml`` that are Ansible tasks which can be executed.
```
mitchnegus:
```suggestion
This takes the form of either an ``INSTALL`` directory with a ``vars.yml`` and a ``tasks.yml`` that are Ansible tasks which can be executed.
```

src/firewheel/cli/configure_firewheel.py
```diff
@@ -178,6 +210,29 @@ def define_set_parser(self) -> argparse.ArgumentParser:
         )
         return parser
 
+    def _update_nested_dict(self, original: dict, updates: dict) -> dict:
+        """
+        This function recursively updates the original dictionary with values
+        from the updates dictionary. If a key in the updates dictionary is a
```
mitchnegus: Maybe we use code syntax since `updates` is a specific keyword argument?

```suggestion
        from the updates dictionary. If a key in the ``updates`` dictionary is a
```

gregjacobus (approved):

PR #167, opened 2025-09-10
feat: Add a CLI command for getting the configuration path

mitchnegus:
This adds a CLI command for getting the path to the FIREWHEEL configuration file.

Example: 
```bash
firewheel config path
> /opt/firewheel/src/firewheel/firewheel.yaml
```


sdelliot (approved):
It all works as expected for me.

PR #187, opened 2025-09-29
fix: Using a build argument to define which FIREWHEEL package to install from GitHub Ref

### LINKED ISSUE #185: Build containers for the active ref
Docker containers published to the container registry [currently install FIREWHEEL from PyPI](https://github.com/sandialabs/firewheel/blob/477b257215833a6503d768e9bc72ca5d8faf6707/docker/firewheel.dockerfile#L70). This means that regardless of the ref, only the latest release of FIREWHEEL will be installed (not any changes between the latest release and the current ref).

The Dockerfile should be updated to pip install FIREWHEEL using the current ref.

Ideally, containers for released versions (and only releases) will still install FIREWHEEL from PyPI to provide broader guarantees of stability.

sdelliot:
# fix: Using a build argument to define which FIREWHEEL package to install from GitHub Ref

## Description
As stated in #185, Docker containers published to the container registry [currently install FIREWHEEL from PyPI](https://github.com/sandialabs/firewheel/blob/477b257215833a6503d768e9bc72ca5d8faf6707/docker/firewheel.dockerfile#L70). This means that regardless of the ref, only the latest _release_ of FIREWHEEL will be installed (not any changes between the latest release and the current ref). This PR addresses this issue by adding a new Docker build argument that allows the GitHub CI to install the FIREWHEEL package from the latest reference.

However, the one big question with this PR is if the project desires to install the release versions from PyPI (as suggested in #185). As implemented, this would not be the case and all docker containers would install from GitHub. However, it can be easily modified with the following:

```yaml
build-args: |
      ${{ github.event_name == 'push' &&  format('FW_PACKAGE_SRC=git+https://github.com/sandialabs/firewheel@{0}', github.ref) || '' }}
```

There are potential issues with this alternative approach though. Without adding explicit dependency on the "Python Publish" workflow, it still may be a race condition if the new package is published before it has to be installed within the docker container. The implemented approach leaves no such ambiguity.

## Related Issue
Closes #185. 

## Type of Change
- [x] Bugfix
- [ ] New feature
- [ ] Documentation update

## Checklist
- ~~[ ] I have commented my code, particularly in hard-to-understand areas.~~
- ~~[ ] I have made corresponding changes to the documentation.~~
- [ ] I have tested my code.


sdelliot:
> I think I agree -- is there actually a need to install from pypi at all? There's no actual functionality difference, right?

I don't think there is any functional difference. It sounds like we all agree that the PR as currently written should solve our issue.

gregjacobus:
Now that I think about it a bit more, I think this functionality might still be a tad misleading. It seems like instead of FW_PACKAGE_SRC defaulting to `firewheel`, it should default to `.` so that any changes in your working directory get built into the container.

I'm imagining the workflow where I want to make some FIREWHEEL changes and build the docker container to test them out. Currently, this PR does not support that workflow, since it'll still install from pypi if a ref isn't given. Or, is the argument that if I'm doing that, I'll override the FW_PACKAGE_SRC variable anyway?

sdelliot:
> Now that I think about it a bit more, I think this functionality might still be a tad misleading. It seems like instead of FW_PACKAGE_SRC defaulting to `firewheel`, it should default to `.` so that any changes in your working directory get built into the container.
> 
> I'm imagining the workflow where I want to make some FIREWHEEL changes and build the docker container to test them out. Currently, this PR does not support that workflow, since it'll still install from pypi if a ref isn't given. Or, is the argument that if I'm doing that, I'll override the FW_PACKAGE_SRC variable anyway?

I agree that it's still a bit misleading though because users that clone the repository likely will expect that the docker container will be built with the associated code in that repository.
Originally, I was thinking that if users are working locally to test changes in their docker container, it would be fairly easy to override ``FW_PACKAGE_SRC``. However, depending on how the container is built, the user may need `.` or `..`, but for users that didn't clone the repo, this would just lead to an error. In your use-case, you could even override the variable to point to your fork/branch and not even have the source code locally on the system using the docker container.

None of what I said really addresses the confusion, but I wonder if updating the README for the docker file would be the right place for this. Maybe a warning banner about it pulling from PyPI unless the variable is set?

mitchnegus:
> > Now that I think about it a bit more, I think this functionality might still be a tad misleading. It seems like instead of FW_PACKAGE_SRC defaulting to `firewheel`, it should default to `.` so that any changes in your working directory get built into the container.
> > I'm imagining the workflow where I want to make some FIREWHEEL changes and build the docker container to test them out. Currently, this PR does not support that workflow, since it'll still install from pypi if a ref isn't given. Or, is the argument that if I'm doing that, I'll override the FW_PACKAGE_SRC variable anyway?
> 
> I agree that it's still a bit misleading though because users that clone the repository likely will expect that the docker container will be built with the associated code in that repository. Originally, I was thinking that if users are working locally to test changes in their docker container, it would be fairly easy to override `FW_PACKAGE_SRC`. However, depending on how the container is built, the user may need `.` or `..`, but for users that didn't clone the repo, this would just lead to an error. In your use-case, you could even override the variable to point to your fork/branch and not even have the source code locally on the system using the docker container.
> 
> None of what I said really addresses the confusion, but I wonder if updating the README for the docker file would be the right place for this. Maybe a warning banner about it pulling from PyPI unless the variable is set?

It seems to me like we should determine who the Dockerfile is for in order to decide the "best" way to handle it—is the Dockerfile mainly for developers who have the repo already? Then it seems like a default which is the local copy and PyPI fallback would be preferable. If it's mainly for users who will just be trying to call out to GitHub to easily get the Dockerfile, then PyPI is preferable. (I ask this because it sounds to me like Greg is suggesting the former and Steven the latter.)

In either case, does it make sense to have a `FW_MC_SRC` variable as well? If we're allowing arbitrary choices of FIREWHEEL, it seems like we should allow arbitrary choices of MC repos (especially because it seems like one of the motivations of testing FW upgrades could be to see how they interact with MC repo upgrades).

PR #202, opened 2025-10-02
feat: Save/export a running experiment

sdelliot:
# Save/export a running experiment

## Description
This draft PR is the start of a conversation around enabling users to save off (or export) a FIREWHEEL experiment. This export can then be loaded back onto the same server or potentially a different server. This solves two primary issues:
1. Rebooting servers during a FIREWHEEL experiment causes issues, but leaving a server running over the weekend is often not necessary and expensive.
2. Enables sharing a running an experiment with others.

There are a few open questions that should be discussed:

### Saving VM Schedules
The PR as written, will save all VMs experiment schedule and then loads that schedule back. Generally, this sounds great. However, there are some nuances with FIREWHEEL which make this difficult. First, the entire schedule is collected, not just the portion of the schedule that has not yet run. That means that upon load, the entire schedule will proceed to run again. After an offline discussion this PR uses this method:

We "intelligently" save off the schedule. That is, we save the experiment time, and remove all scheduled events prior to the "current time". In this case, because negative time is shown as per-VM, no negative time events would be saved. That is, we explicitly **prohibit** saving an experiment in negative time! Rather, only positive time events that "have not yet run" will be saved. 

A discussed, though chosen for a later

**Option 3:** We can potentially update the VM resource handler to identify duplicate entries or entries that have happened before. To me, this seems very risky and could break existing expectations of the system.

### Issues with minimega
This feature relies on minimega v3.0.0!

## Type of Change
- [ ] Bugfix
- [x] New feature
- [x] Documentation update

## Checklist
- [x] I have tested my code.


docs/source/tutorials/save_load.rst
```diff
@@ -0,0 +1,425 @@
...
+****************
+
+In this tutorial, you will first create and save a known-good checkpoint of a
+running experiment. After saving, the current experiment is paused, which gives
+you a natural point to either stop or resume and continue working from that
+state.
+
+The first diagram below shows that :ref:`helper_save` workflow.
+
+.. graphviz::
```
gregjacobus, line 54: this and the other graph don't seem to be building correctly

src/firewheel/cli/helpers/save
```diff
@@ -0,0 +1,741 @@
...
+            f'Default: "{default_experiment_name}".'
+        ),
+    )
+    parser.add_argument(
+        "-c",
+        "--complete",
+        action="store_true",
+        help=(
+            "If provided, save all files from the experiment, including cached "
+            "images and VM resources."
```
gregjacobus, line 815: presumably this is useful for copying experiments between machines? maybe add a blurb about that?

docs/source/tutorials/save_load.rst
```diff
@@ -0,0 +1,425 @@
...
+************************
+
+For this tutorial, we will use the :ref:`router-tree-tutorial` experiment because
+it is small, familiar, and provides accessible Ubuntu VMs for verification.
+
+Launch the experiment with:
+
+.. code-block:: bash
+
+    $ firewheel experiment tests.router_tree:3 minimega.launch
```
gregjacobus, line 183: control network isn't getting loaded properly in my tests after save

gregjacobus (changes requested):

gregjacobus (changes requested):
I added the following line as part of `tests.vm_gen`:

```
vm.run_executable(1, "touch", "/tmp/FLAG") 
```

I ran `firewheel save` in the 60 second window between negative and positive time. Then did `firewheel load`. But `touch` command was never executed

docs/source/tutorials/save_load.rst
sdelliot: Fixed in [b18c334](https://github.com/sandialabs/firewheel/pull/202/commits/b18c33407a9e0eb37aaafc2f5a3422ee3d3dfeaf)

docs/source/tutorials/save_load.rst
sdelliot: If you are building it on your own system, you need to install graphviz. I thought we installed it on the CI nodes, but it appears not. I wonder if I should just get the image and use that instead of trying to be "fancy".

src/firewheel/cli/helpers/save
```diff
@@ -0,0 +1,760 @@
...
+        )
+
+        for file in schedule_db_paths:
+            try:
+                prune_schedule_entries(
+                    experiment_time["seconds_since_start"],
+                    file,
+                    schedule_dir / file.name,
+                    console,
+                )
```
gregjacobus, line 667: This fixes the problem where (some) negative time vm resources were being run multiple times on a loaded experiment when the save occurred during the 60 second window between configuration and experiment start. The problem is that `experiment_time["seconds_since_start"]` returns a negative value during that window. So if you save early in that window, that value is `-50`. Then when `prune_schedule_entries` runs, a vm resource scheduled at time `-20`, for example, would run again, even though it already ran before the experiment was saved.

```suggestion
                # It is possible for the experiment_time to be negative if the
                # VMs have completed configuring but positive time has not yet
                # been reached. In this case, we want to set the
                # seconds_since_start to 0 so that negative time VMs do not get
                # re-run
                seconds_since_start = max(0, experiment_time["seconds_since_start"])

                prune_schedule_entries(
                    seconds_since_start,
                    file,
                    schedule_dir / file.name,
                    console,
                )
```

src/firewheel/lib/utilities.py
```diff
@@ -28,70 +227,66 @@ def render_rich_string(text):
...
-    """
-    Identify and return the members of a :py:class:`tarfile.TarFile` that are considered safe.
-    See the documentation for :py:meth:`tarfile.TarFile.extractall` for more information.
-    This function, as well as :py:func:`badlink` and :py:func:`badpath` were based on
-    https://stackoverflow.com/a/10077309.
+def get_safe_tarfile_members(
+    tarfile_obj: tarfile.TarFile,
+    base: Path,
+) -> list[tarfile.TarInfo]:
+    """Return tar members considered safe to extract under a base directory.
```
gregjacobus, line 292: Without the suggested change, this function changes the number of arguments to the function but doesn't change the call to it in `lib/minimega/file_store.py`. Defaulting to current directory seems like a reasonable option

```suggestion
def get_safe_tarfile_members(
    tarfile_obj: tarfile.TarFile,
    base: Path = Path("."),
) -> list[tarfile.TarInfo]:
    """Return tar members considered safe to extract under a base directory.
```

gregjacobus (approved):

PR #226, opened 2025-12-23
feat: Better default config values

gregjacobus:
# Better default config values

## Description
A number of the default configuration parameters are pretty useless as is. Setting them to reasonable defaults will make the process of installing FIREWHEEL much simpler.

## Related Issue
N/A

## Type of Change
- [ ] Bugfix
- [x] New feature
- [ ] Documentation update

## Checklist
- [x] I have tested my code.

## Additional Notes
N/A


mitchnegus (approved):
These all seem reasonable to me—`localhost` seems well defined, and I think most any Unix system will have a `/tmp` directory, so I have no objections.

I'll defer to Steven on whether there might be reasons to not assume names based on historical context.

mitchnegus (approved):
These are minor suggestions for the wording. I think they are okay as is, but the wording "left" implies that they default to the empty string (which will no longer be true).

gregjacobus:
@mitchnegus I do agree with all the wording suggestions you made. It turns out, however, that it falling back to `system.default_output_dir` is simply not true, which is why having these not be set is a problem to begin with. I think this PR should change to fixing that bug instead. The only default argument then that would need to stay changed would be `minimega.experiment_interface`. 

I would also prefer to have a way to set `cluster.compute` and `cluster.control` to localhost. Or have a way to set it automatically. Then, a fresh installation of FIREWHEEL could work without touching the config file at all. Not sure if minimega likes having the cluster as localhost though. If anybody has insight on that, it would be greatly appreciated.

PR #233, opened 2026-01-13
feat: Enhance progress outputs for image cache uploads

mitchnegus:
# Enhance progress outputs for image cache uploads

## Description
In service of improving the user experience of launching FIREWHEEL experiments, this MR begins adding more functionality to the `ModelComponent` and `ModelComponentManager` objects.

This MR contains a few substantive changes:
1. Most obvious to the user, the output from loading images to the image cache has been expanded. Images are grouped by MC, and while the load time continues to display independently per-image, the spinner is now associated with a progress output of the images to be loaded for any given MC. Also, large VM resources (over 250 MB) now also receive progress bars during upload.
2. The change described in (1) arose from needing to use a distinct layout of renderables from the rich library to separately display image cache upload progress, experiment graph progress (not implemented in this MR), experiment time progress (also not implemented in this MR), and the outputs of MC plugins. On the backend, that meant compartmentalizing image uploads to be represented by progress bars to match the [live progress](https://github.com/Textualize/rich/blob/master/examples/live_progress.py) example provided by rich, as well as adjusting some of the image handling to know before creating the progress bar whether an image would need updating.
3. In the process of accomplishing (2), the `ModelComponent` object was heavily refactored. It handled VM resources and images using very similar, but entirely separate upload processes. Now, it performs uploads in a uniform fashion, depending on the subtleties of each upload type defined via a new object class—a `_ModelComponentFile`. This object is notionally analagous to the existing `FileStoreFile` object from the `lib.minimega.file_store` module, but this represents the file as part of the model component rather than the file store. Ultimately, it seems like these file-like objects might do well to be combined eventually, but that feels like a separate issue.

## Type of Change
- [ ] Bugfix
- [x] New feature
- [ ] Documentation update

## Checklist
- [x] I have tested my code.

### DIFF
--- a/src/firewheel/control/model_component.py
+++ b/src/firewheel/control/model_component.py
@@ -509,7 +509,7 @@ class ModelComponent:
         Upload all image files from the manifest.
 
         Returns:
-            dict: Actions for each specified file. Possible actions are:
+            dict: Statuses for each specified file. Possible statuses are:
 
             * ``no_date`` -- There was no upload date for the given file in the
                   ``ImageStore``. It was uploaded.
@@ -525,30 +525,38 @@ class ModelComponent:
             An empty dictionary is returned if no images are identified.
 
         """
-        return {
-            image_path: self._upload_image(image_path)
+        image_statuses = {
+            image_path: self._evaluate_image(image_path)
             for image_field in self.manifest.get("images", {})
             for image_path in image_field.get("paths", [])
         }
+        # Filter images that require uploads (and upload them)
+        upload_images = {
+            image_path: status
+            for image_path, status in image_statuses.items()
+            if status in {"no_date", "new_hash"}
+        }
+        if upload_images:
+            for image_path, status in upload_images.items():
+                self._upload_image(image_path, status)
+        return image_statuses
 
-    def _upload_image(self, image_path):
+    def _evaluate_image(self, image_path):
         """
-        Upload an image to the ``ImageStore``.
+        Evaluate an image's status (governing uploads to the ``ImageStore``).
 
         Args:
-            image_path (str): A path to an image to be added to the ``ImageStore``.
+            image_path (str): A path to an image to be evaluated.
 
         Returns:
-            str: An action for the specified image file. Possible actions are:
+            str: A status for the specified image file. Possible statuses are:
 
             * ``no_date`` -- There was no upload date for the given file in the
-                  ``ImageStore``. It was uploaded.
+                  ``ImageStore``.
             * ``new_hash`` -- The modified time of the file on disk differs from the
                   last upload time in the ``ImageStore`` and the hashes did not match.
-                  File was uploaded.
             * ``same_hash`` -- The file on disk was modified after the upload time in
-                  the ``ImageStore`` but the hashes are the same. File was
-                  not uploaded.
+                  the ``ImageStore`` but the hashes are the same.
             * :py:data:`False` -- None of the other conditions occurred. For example,
                   the file on disk was modified before the ``ImageStore`` upload time.
         """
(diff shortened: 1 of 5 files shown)


src/firewheel/control/model_component.py
```diff
@@ -515,89 +533,130 @@ def _upload_images(self):
...
+            * ``no_date`` -- There was no upload date for the given file in the
+                  ``ImageStore``. It was uploaded.
+            * ``new_hash`` -- The modified time of the file on disk differs from the
+                  last upload time in the ``ImageStore`` and the hashes did not match.
+                  File was uploaded.
+            * ``same_hash`` -- The file on disk was modified after the upload time in
+                  the ``ImageStore`` but the hashes are the same. File was
+                  not uploaded.
+            * :py:data:`False` -- None of the other conditions occurred. For example,
+                  the file on disk was modified before the ``ImageStore`` upload time.
```
sdelliot:
```suggestion
                  the file on disk was modified before the :py:class:`ImageStore <firewheel.control.image_store.ImageStore>` upload time.
```

src/firewheel/control/model_component.py
```diff
@@ -430,7 +448,7 @@ def _upload_vm_resource(self, resource):
             self.vm_resource_store.add_file(path)
             return "no_date"
 
-        if last_modified_date != upload_date:
+        if modification_timestamp != upload_date:
             self.log.debug("Resource on disk is different from store. Checksuming.")
```
sdelliot:
```suggestion
            self.log.debug("Resource on disk is different from store. Checksumming.")
```

sdelliot (changes requested):

src/firewheel/control/model_component.py
```diff
@@ -515,89 +533,130 @@ def _upload_images(self):
...
+            return "same_hash" if disk_hash == store_hash else "new_hash"
+        return False
+
+    def _determine_image_actions(self, image_statuses):
+        # Determine (and filter) images that require uploads
+        return {
+            image_path: "Adding" if status == "no_date" else "Updating"
+            for image_path, status in image_statuses.items()
+            if status in {"no_date", "new_hash"}
+        }
```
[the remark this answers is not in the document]
mitchnegus: Ok, so I decided that this behavior was directly related to the "super gross way of handling this return value" from above. Towards that end, I added a new set of objects (types of `_ModelComponentFile`) and the enum you suggested. 

This let me eliminate this construct entirely, and unify almost all of the behavior of images and VM resources, which were previously handled separately. The best part about this is that now the `ModelComponent` can do progress for VMR uploads as well as images, which I believe was a feature requested out-of-band.

mitchnegus:
@sdelliot I mentioned it in response to your comments above, but I cleaned up the module a fair bit. The issues you rose were all tied to that weird return value construct, and I found that the easiest way to clean things up was to use your enum suggestion paired with a couple new objects representing the images and VM resources.

There are still 4 tests that fail. I held off on fixing these because I wanted suggestions on solving them. Previously, the `ModelComponent` object would follow this logic:

1. No image in file store, upload it.
2. If it exists, compare times. Replace file store images with newer MC images that are not duplicates.

This left some undefined behavior in other cases—such as when an MC image was older than the file store image... and my understanding is that we actually just (silently) replaced the newer file store image in this case. This seems to be what the currently failing test cases are concerned with.

My current draft of this takes the opposite approach, and rather than deleting the newer file store image, it raises a user warning that the cache is out of sync. The user can then decide to blow away the cache, or investigate, or whatever. This seems like a pretty weird case that only developers might ever actually be _intendng_ to reach, but even if the best solution here is almost always to just reset the file, that feels like it's generally the wrong solution to force.

Changing the logic from above, the upload now follows (`model_component.py`, line 89):
1. No image in the store, upload it.
2. If it exists, compare hashes of MC file and the file store file. Ignore files with matching hashes.
4. If the hashes differ, and the file store is out of date, reupload.
5. If the hashes differ but the file store is newer, do not reupload but give a warning.

This feels like a better approximation of the true logic, but I'm still testing to see if the hash check coming first will be problematically slow for large VMs.

sdelliot:
> @sdelliot I mentioned it in response to your comments above, but I cleaned up the module a fair bit. The issues you rose were all tied to that weird return value construct, and I found that the easiest way to clean things up was to use your enum suggestion paired with a couple new objects representing the images and VM resources.
> 
> There are still 4 tests that fail. I held off on fixing these because I wanted suggestions on solving them. The `ModelComponent` object follows this logic:
> 
>     1. No image in file store, upload it.
> 
>     2. If it exists, compare times. Replace file store images with newer MC images that are not duplicates.
> 
> 
> This leaves some undefined behavior in other cases—such as when an MC image was older than the file store image... One of the currently failing test cases suggest that this might happen if the user were to copy the MC version of an image elsewhere, then replace it and reupload it (maybe for debugging something), and finally reverting it back to the original and reuploading that again. Previously, we seemed to just (silently) replace the newer file store image in this case.
> 
> My current draft of this takes an opposite approach, and rather than deleting the newer file store image, it raises a user warning that the cache is out of sync. The user can then decide to blow away the cache, or investigate, or whatever. Getting into this state seems like a pretty weird case that only developers might ever actually be _intendng_ to reach, but even if the best solution here for most users is almost always to just reset the cache, silently deleting files feels to me like a big no-no.
> 
> If that feels like the right approach, I can add, update, and/or remove tests to match. Otherwise, I'm open to other ideas.
> 
> **Side note:** A cool benefit of this round of refactoring is that VM resources can now also easily receive progress bars. I set it so that large VMRs (over 250 MB) will get them when they're uploaded to the cache. I think that feature had been requested out-of-band.

That is certainly one of the reasons why we had that code in there originally. The second is for the case of using `vm builder` pointing to the image in the cache. In the case that you accidentally modify the image, you wouldn't want that for your experiment. 

I agree that it doesn't seem valuable or wise to just silently delete the newer images and I am on board with your approach. If a user is messing around with the images that much, they likely wouldn't want their image deleted and maybe would want to save it off. 

The new Enum structure really cleans up the code a lot! I also agree that we can likely merge the FileStoreFile and _ModelComponentFile, but I am okay with punting that down the road.

PR #243, opened 2026-02-19
feat: ADB Driver

gregjacobus:
# Pull Request Title

## Description
This primarily adds a new driver for ADB (android debug bridge) to facilitate emulating Android devices in FIREWHEEL

## Type of Change
- [ ] Bugfix
- [x] New feature
- [ ] Documentation update

## Checklist
- [ ] This PR conforms to the process detailed in the [Contributing Guide](https://sandialabs.github.io/firewheel/developer/contributing.html).  
- [ ] I have included no proprietary/sensitive information in my code. 
- [ ] I have performed a self-review of my code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have tested my code.

## Additional Notes
Please provide any additional information or context for your pull request here.


src/firewheel/vm_resource_manager/drivers/adb_driver.py
```diff
@@ -0,0 +1,1090 @@
...
+    # Connection / lifecycle
+    # -------------------------------------------------------------------------
+
+    def _wait_for_device_online(self):
+        """Wait until the ADB device is online and shell commands work."""
+        while not self.ping():
+            self.log.debug(
+                "Waiting for Android device %s to come online", self.adb_name
+            )
+            time.sleep(1)
```
mitchnegus: Do we want a max attempts or retry limit here to prevent infinite loops? If so, you may have a better idea than I do about what sensible limits would be.

src/firewheel/vm_resource_manager/drivers/adb_driver.py
```diff
@@ -0,0 +1,1090 @@
...
+
+    def get_time(self):
+        """
+        Get current Android guest time.
+
+        Returns:
+            float: Seconds since epoch.
+        """
+        output = self._shell("date +%s").strip()
+        return float(int(output.splitlines()[-1]))
```
mitchnegus: Not sure if this is captured elsewhere higher in the traceback, but might be good to add a quick error here to save time debugging if `_shell` returns empty. 

Maybe something like:

```suggestion
        if not (output := self._shell("date +%s")):
            raise RuntimeError("Failed to acquire date")
        final_line = output.strip().splitlines()[-1]
        return float(int(final_line))
         
```

src/firewheel/vm_resource_manager/drivers/adb_driver.py
```diff
@@ -0,0 +1,1090 @@
...
+        schedule_entry.reboot_file = schedule_entry.working_dir / "reboot"
+        schedule_entry.call_args_filename = (
+            schedule_entry.working_dir / "call_arguments.sh"
+        )
+
+        call_arguments = (
+            f"#!{self.ANDROID_SHELL}\n"
+            'CURRENT_DIR="$(dirname "$0")"\n'
+            f"cd {schedule_entry.working_dir}\n"
+            f"{schedule_entry.exec_path!s}"
```
mitchnegus: Could/should either/both of these paths get quoting? (I guess matters only if these might have spaces, which I'm guessing is exceedingly unlikely... but maybe still worth being safe)

```suggestion
            f'cd "{schedule_entry.working_dir}"\n'
            f'"{schedule_entry.exec_path!s}"'
```

(Note: Looks like this may also be a pattern [derived from the `AbstractDriver`](https://github.com/sandialabs/firewheel/blob/964d94fd8eb669e95df597348f8c92de432eb397/src/firewheel/vm_resource_manager/abstract_driver.py#L797))

mitchnegus (dismissed):
Changes look good since the last time I checked. I gave it a quick AI pass and it surfaced the following suggestions (among others I triaged). 

None are showstoppers, so I approve this MR. If you feel free to accept them I'll approve again, otherwise we could make issues for them separately (or just ignore them if the code paths in question aren't worth checking).

src/firewheel/vm_resource_manager/drivers/adb_driver.py
sdelliot: I _think_ that an infinite loop is the way to go here just because it's a bit different than the QGA connection thing. I also am not sure on timeout values? I think this is something we can revisit when we have more data.

src/firewheel/vm_resource_manager/drivers/adb_driver.py
sdelliot: I like the idea, but I changed the approach.

src/firewheel/vm_resource_manager/drivers/adb_driver.py
```diff
@@ -0,0 +1,1093 @@
...
+        Get current Android guest time.
+
+        Returns:
+            float: Seconds since epoch.
+        """
+        if not (output := self._shell("date +%s")):
+            raise RuntimeError("Failed to acquire date")
+        final_line = output.strip().splitlines()[-1]
+        return float(int(final_line))
+
```
mitchnegus: My bad; extra line added in commit suggestion 🙃 
```suggestion
```

mitchnegus (dismissed):

mitchnegus (approved):

sdelliot (approved):

PR #173, opened 2025-09-12
ci: Adding workflow_call to make a consistent ecosystem for FIREWHEEL

sdelliot:

PR #254, opened 2026-03-24
ci: Transition FIREWHEEL to using the CI repo for all shared capabilities

mitchnegus:
# Transition FIREWHEEL to using the CI repo for all shared capabilities

## Description

This extracts all functionality from this FIREWHEEL package repo that is now covered by the [FIREWHEEL CI repo](https://github.com/sandialabs/firewheel_ci). Changes here largely follow those in the MC repos (e.g., those in sandialabs/firewheel_repo_base#7, sandialabs/firewheel_repo_linux#7, and similar PRs in other MC repos).

- Composite actions have been removed because they are now entirely contained in the CI repo.
- Workflow files have been reduced to simply call reusable workflows in the CI repo.

**Also, this removes the need for major version tags (and includes removal of the corresponding workflow).**

Relatedly, this likely closes #173. 

## Type of Change
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation update
- [x] Other (please describe): CI refactor

## Checklist
- [ ] ~I have made corresponding changes to the documentation.~ _N/A_
- [x] I have tested my code.


.github/workflows/python-publish.yml
```diff
@@ -1,44 +1,12 @@
...
-    - name: Build package
-      run: python -m build
-    - name: Publish package
-      uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
-      with:
-        user: __token__
-        password: ${{ secrets.PYPI_API_TOKEN }}
+  call-firewheel-publishing:
+    name: Publish FIREWHEEL Packages
+    uses: sandialabs/firewheel_ci/.github/workflows/python-publish.yml@v1
```
sdelliot, line 12: Do we need to the following (like https://github.com/sandialabs/firewheel_repo_base/blob/main/.github/workflows/python-publish.yml):


```suggestion
    uses: sandialabs/firewheel_ci/.github/workflows/python-publish.yml@v1
    secrets:
      PYPI_API_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
```

.github/workflows/testing.yml
```diff
@@ -29,7 +29,10 @@ jobs:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 0  # full history
+        fetch-tags: true
```
sdelliot, line 35: Is there a compelling benefit to do this?

sdelliot (changes requested):

.github/workflows/testing.yml
mitchnegus: Yes. I noticed in one of the actions that FIREWHEEL was reporting that it was installing v0.0.1... in the CI pipeline. I discovered that the CI pipeline installs FIREWHEEL at the current commit, but because tags are not fetched and the depth is unset (therefore no commits other than that very last one are fetched either), the setuptools SCM package has no versioning info. It thus reverts to some pre-v0.0.1 tag. 

An example of the old behavior can be seen in the old workflows: see [line 220 on this action](https://github.com/sandialabs/firewheel/actions/runs/23508312383/job/68422193470#step:7:223) from an earlier commit in this PR. New behavior is [here](https://github.com/sandialabs/firewheel/actions/runs/23509437662/job/68426161233#step:7:223).

I don't know that this actually has any bearing on the CI job being run, but it seemed like it might be confusing and it was easy enough to solve. (Credit where credit is due, some AI bot figured this issue out with the solution in like 2 prompts.) Also, since we don't store large binaries with FIREWHEEL, it seemed to have minimal downsides—checking out a full-but-not-large repo takes negligible time compared to everything else we do.

.github/workflows/python-publish.yml
mitchnegus: Most likely? (I'll check on this... it's likely an oversight)

sdelliot (dismissed):

sdelliot (approved):
LGTM

PR #283, opened 2026-05-29
fix: atomic grpc operations

### LINKED ISSUE #275: [BUG] Non-atomic gRPC experiment start time
### Description
Experiment start time initialization is not atomic in the gRPC timing service. Multiple VM Resource Handlers can race to set the experiment start time when they transition to `VMState.CONFIGURED` at nearly the same time.

The current flow checks whether a start time already exists and, if not, writes a new one:
- `VMResourceHandler.set_state()` calls `api.add_experiment_start_time()` when `not_ready_count == 0`
- `ExperimentStart.add_start_time()` does a read-then-write:
  - `get_start_time()`
  - if unset, `set_experiment_start_time(new_time)`

Because this is not atomic, multiple handlers can observe "no start time" and both write competing values.

### Steps to Reproduce
1. Launch an experiment with multiple VMs that finish negative-time configuration very close together.
2. Ensure multiple VMs can transition to `VMState.CONFIGURED` within a narrow timing window.
3. Observe concurrent calls to `api.add_experiment_start_time()` from different handlers.
4. Inspect the gRPC server state or logs for multiple appended values in `experiment_start_times`.

### Expected Behavior
Experiment start time should be initialized exactly once using an atomic "set-if-absent" operation. All handlers should observe the same single start time.

### Actual Behavior
Multiple handlers may race and append multiple start times to `self.dbs[db]["experiment_start_times"]`. The server returns the first value, but multiple writes can still occur, making initialization non-atomic and timing-dependent.

### Environment
- **Operating System**: Ubuntu
- **Version**: 2.12
- **Any Model Components Impacted**: `vm_resource_manager`, `firewheel.lib.grpc`

### Additional Context
Relevant code paths:
- `firewheel/vm_resource_manager/vm_resource_handler.py`
  - `VMResourceHandler.set_state()`
- `firewheel/vm_resource_manager/experiment_start.py`
  - `ExperimentStart.add_start_time()`
- `firewheel/lib/grpc/firewheel_grpc_server.py`
  - `SetExperimentStartTime()`

Current server implementation appends every request:
```python
def SetExperimentStartTime(self, request, context):
    db = request.db
    self.dbs[db]["experiment_start_times"].append(request)
    return self.dbs[db]["experiment_start_times"][0]
```

Suggested fix:
- Add a new atomic gRPC RPC such as `InitializeExperimentStartTimeIfAbsent`
- Or change `SetExperimentStartTime` semantics server-side to only initialize when no start time exists

### Checklist
- [x] I have included no proprietary/sensitive information in my issue.

sdelliot:
## Description
This PR fixes a race in experiment start time initialization by making `SetExperimentStartTime` atomic at the gRPC server and updating the client-side `ExperimentStart` logic to rely on that atomic server behavior.

Previously, the client performed a read-then-write sequence when establishing the experiment start time. If multiple VM Resource Handlers reached `VMState.CONFIGURED` at nearly the same time, they could race:
1. both observe that no start time exists
2. both attempt to write a start time
3. the server would append both values

This PR changes the implementation so that:
- `SetExperimentStartTime` atomically initializes the experiment start time if absent
- if the start time is already set, the existing value is returned unchanged
- `ExperimentStart.add_start_time()` no longer does a client-side pre-read before writing
- timing-related state in the gRPC server is guarded by locking to avoid concurrent access issues

This change is required to eliminate a race in the experiment start-time synchronization mechanism.

The previous implementation relied on client-side synchronization:
- `ExperimentStart.add_start_time()` called `get_start_time()`
- if the result was `None`, it computed a proposed start time and called `set_experiment_start_time()`

That pattern is not atomic and allowed multiple handlers to race during experiment startup. On the server side, `SetExperimentStartTime()` appended every requested start time and returned the first one, which meant multiple competing writes could still occur.

This PR solves that problem by moving the synchronization responsibility to the gRPC server:
- `SetExperimentStartTime()` now acts as an atomic initialize-if-absent operation
- `GetExperimentStartTime()`, `SetExperimentLaunchTime()`, `GetExperimentLaunchTime()`, and `InitializeExperimentStartTime()` are protected with the same server-side lock for timing-state consistency
- `ExperimentStart.add_start_time()` now directly calls `set_experiment_start_time()` and returns the stored value from the server

This keeps the API surface unchanged while making start-time establishment correct under concurrency.

Resolves https://github.com/sandialabs/firewheel/issues/275.

## Type of Change ##
- Bugfix

## Checklist ##

## Additional Notes ##
There is no proto/RPC shape change in this PR. Instead, the semantics of the existing `SetExperimentStartTime` RPC are tightened so that the operation is atomic and idempotent from the client perspective.

This approach was chosen to:
- keep the change narrowly focused
- avoid introducing a second overlapping timing RPC
- ensure all existing callers benefit automatically without changing the public API surface


src/firewheel/lib/grpc/firewheel_grpc_server.py
```diff
@@ -58,70 +57,32 @@ def __init__(self):
...
+        # access/mutation is possible.
+        self._db_lock = threading.RLock()
+
         self.dbs = {}
         for db in ["test", "prod"]:
             os.makedirs(os.path.join(self.cache_dir, db), exist_ok=True)
             self._init_db(db)
 
-    def _read_repository_db_from_file(self, db):
-        """
```
sdelliot, line 67: Note that this function was lingering from #48 and should be removed.

sdelliot (commented):
Confirmed human code review and testing.

PR will be in draft mode until changes made in [49261a3](https://github.com/sandialabs/firewheel/pull/283/commits/49261a3928af43944ebd1161a6016c1a375b9fea) have been tested

src/firewheel/lib/grpc/firewheel_grpc_server.py
```diff
@@ -211,18 +180,19 @@ def GetExperimentStartTime(self, request, context):  # noqa: N802,ARG002
...
-            error_details = "IndexError. No start time available yet."
-            error_code = grpc.StatusCode.OUT_OF_RANGE
-            context.abort(code=error_code, details=error_details)
+        with self._db_lock:
+            with contextlib.suppress(IndexError):
+                return self.dbs[db]["experiment_start_time"]
+
+        error_details = "IndexError. No start time available yet."
+        error_code = grpc.StatusCode.OUT_OF_RANGE
+        context.abort(code=error_code, details=error_details)
```
mitchnegus: I assume that this was just kept as backwards compatibility, but without the value keyed to `"experiment_start_times"` being a list anymore, this should no longer have the possibility of raising an `IndexError`. 

```suggestion
            return self.dbs[db]["experiment_start_time"]
```

(I believe the `return None` below would also then become redundant/extraneous/unreachable and so also should probably be removed.)

src/firewheel/lib/grpc/firewheel_grpc_server.py
```diff
@@ -232,10 +202,11 @@ def InitializeExperimentStartTime(self, request, context):  # noqa: N802,ARG002
...
         db = request.db
-        with contextlib.suppress(KeyError):
-            self.dbs[db]["experiment_launch_time"] = None
-        with contextlib.suppress(KeyError):
-            self.dbs[db]["experiment_start_times"] = []
+        with self._db_lock:
+            with contextlib.suppress(KeyError):
+                self.dbs[db]["experiment_launch_time"] = None
+            with contextlib.suppress(KeyError):
+                self.dbs[db]["experiment_start_time"] = None
```
mitchnegus: I'm also unclear on the purpose of these `contextlib.suppress(KeyError)` guards.

- If `self.dbs[db]` exists, then it sets the value regardless of whether or not `"experiment_launch_time"` or `"experiment_start_time"` exist. (So no `KeyError` is ever thrown.)
- If `self.dbs[db]` does not exist, that seems like something we should probably tell the user, like "Hey, you're assigning to some DB that doesn't actually exist."

Are there grpc reasons that these have to happen, like if `self.dbs[db]` may not exist yet and we're just trying until it does? (And if so, can we handle that more explicitly/clearly?)

src/firewheel/lib/grpc/firewheel_grpc_server.py
```diff
@@ -292,10 +264,20 @@ def SetVMTimeByUUID(self, request, context):  # noqa: N802,ARG002
...
+                ].current_time = request.current_time
+                return copy.deepcopy(self.dbs[db]["vm_mappings"][request.server_uuid])
+            except KeyError:
+                error_details = (
+                    f"IndexError. No vm_mapping found for {request.server_uuid}"
+                )
+                error_code = grpc.StatusCode.OUT_OF_RANGE
+                context.abort(code=error_code, details=error_details)
+
+        return None
```
mitchnegus: At risk of changing the interface's behavior, should this error reflect that it is hitting a `KeyError` not an `IndexError`? (Maybe it's not actually changing the interface behavior, since I see no previous error handling in this method...)

Separately, there are a number of lookups in this chunk that could raise `KeyError` without it actually being the `request.server_uuid` lookup. I think using the `get_vm_mapping` method (same is used by `GetVMMappingByUUID`) would solve that:

```suggestion
        with self._db_lock:
            vm_mapping = self.get_vm_mapping(
                self.dbs[db]["vm_mappings"], request.server_uuid
            )
            if vm_mapping:
                vm_mapping.current_time = request.current_time
                return copy.deepcopy(vm_mapping)
            else:
                error_details = (
                    f"IndexError. No vm_mapping found for {request.server_uuid}"
                )
                error_code = grpc.StatusCode.OUT_OF_RANGE
                context.abort(code=error_code, details=error_details)

        return None
```

(And/or, that but without the reference to `IndexError`)

src/firewheel/lib/grpc/firewheel_grpc_server.py
```diff
@@ -322,16 +304,19 @@ def SetVMStateByUUID(self, request, context):  # noqa: N802,ARG002
...
+                self._update_not_ready_vmms(request, db)
+                vmm = self.dbs[db]["vm_mappings"][request.server_uuid]
+                vmm.state = request.state
+                return copy.deepcopy(vmm)
+            except KeyError:
+                error_details = (
+                    f"IndexError. No vm_mapping found for {request.server_uuid}"
+                )
+                error_code = grpc.StatusCode.OUT_OF_RANGE
+                context.abort(code=error_code, details=error_details)
```
mitchnegus: Same comment as above:

```suggestion
        with self._db_lock:
            db = request.db
            self._update_not_ready_vmms(request, db)
            vm_mapping = self.get_vm_mapping(
                self.dbs[db]["vm_mappings"], request.server_uuid
            )
            if vm_mapping:
                vm_mapping.state = request.state
                return copy.deepcopy(vm_mapping)
            else:
                error_details = (
                    f"IndexError. No vm_mapping found for {request.server_uuid}"
                )
                error_code = grpc.StatusCode.OUT_OF_RANGE
                context.abort(code=error_code, details=error_details)
```

mitchnegus (commented):
Most of my questions on a first read-through stem from the `firewheel_grpc_server.py` module's use of the `contextlib.suppress(KeyError)` lines. If there's a motivation for why we rely so heavily on that based on how this is used, feel free to let me know and I can reevaluate.

src/firewheel/lib/grpc/firewheel_grpc_server.py
mitchnegus: After looking at a couple of these, it seems like a method like the following would consolidate most...

```python
def _get_server_specific_vm_mapping(self, context, request):
    vm_mapping = self.get_vm_mapping(self.dbs[db]["vm_mappings"], request.server_uuid)
    if vm_mapping is None:
        context.abort(
            code=grpc.StatusCode.OUT_OF_RANGE,
            details=f"IndexError. No vm_mapping found for {request.server_uuid}",
        )
    return vm_mapping
```

Then these would all look more like:

```suggestion
        with self._db_lock:
            db = request.db
            self._update_not_ready_vmms(request, db)
            vm_mapping = self._get_server_specific_vm_mapping(context, request)
            vm_mapping.state = request.state
            return copy.deepcopy(vm_mapping)
```

PR #285, opened 2026-05-29
deps: Update ruff to 0.15

sdelliot:
## Description
Related to #274 but with the necessary fixes. This should also reduce the dependabot noise for ruff.

## Type of Change ##

- deps

## Checklist ##


mitchnegus:
Looks to me like the current errors are because we are still [running Flake8 via tox](https://github.com/sandialabs/firewheel/blob/441ae2112ce209869f65373d441a9be3239f0e43/tox.ini#L100). Support for these codes (`DOC109`, `DOC110`) seem to [still be under development in ruff](https://github.com/astral-sh/ruff/issues/12434). 

However, it also seems like the official Google docstring style guide [says](https://google.github.io/styleguide/pyguide.html#doc-function-args) that these docstring-based type hints are not required if the type hints are provided in the code.

Maybe we just go through those two files failing the tests (`.../lib/minimega/file_store.py` and `.../lib/utilities.py`) and ensure that all of the function signatures have type hints and all docstrings do not, and then we add Flake8 per-file-ignores for `DOC109` and `DOC110` in those two files.

For example, in `tox.ini`, something like:

```ini
[flake8]
...

per-file-ignores =
    # Ignore missing docstring type hints
    src/firewheel/lib/utilities.py: DOC109, DOC110
    src/firewheel/lib/minimega/file_store.py: DOC109, DOC110
```

(I can make an MR to your branch if you agree with that suggestion, or if you beat me to a fix.)

mitchnegus (approved):

PR #288, opened 2026-06-09
feat: Add Python 3.14 compatibility

mitchnegus:
# feat: Add Python 3.14 compatibility

(Pardon the AI language style in the MR, but generating it this way from the patch files is so much faster...)

## Description
This MR updates FIREWHEEL to support Python 3.14. It primarily addresses multiprocessing compatibility (due to updates to picklability of objects). The changes address Python 3.14 pickling behavior by moving multiprocessing target functions out of local scope and into module-level functions so they can be safely serialized. In particular:

- `ExperimentGraph` shortest-path worker logic was refactored to use a top-level helper function and stable graph node IDs instead of locally scoped closures.
- `minimegaAPI._check_version()` was updated to use a top-level worker target and pickle-safe arguments for compatibility checks.
- `ExperimentGraph` multiprocessing usage was updated to use an explicit multiprocessing context for queue and process creation.
- Additional robustness was added around spawned process lifecycle management through a reusable context manager that ensures cleanup of processes and queues.
- `minimega` API code was further improved with type annotations and some small cleanup/refactoring for readability and maintainability. (To pass linting checks.)
- Docstrings were cleaned up for consistency.
- Tests and project metadata were updated to indicate and validate support for Python 3.14.

Python 3.14 changes multiprocessing/pickling behavior in a way that breaks some existing patterns used in FIREWHEEL, specifically locally defined functions used as multiprocessing targets. (See the commentary on [forkserver restrictions](https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods) in the [3.14 Changelog](https://docs.python.org/3/whatsnew/3.14.html#concurrent-futures). It is in the section on concurrent futures, which may be a better way to handle this pattern eventually, at least in the `minimega/api` module.) 

These changes caused compatibility issues in code paths that spawn child processes for graph traversal and minimega binding/version validation.

This MR resolves that problem by:
- Replacing non-pickleable local worker functions with top-level helpers
- Passing only pickleable data into child processes
- Making process/queue management more robust

Other additions for Python 3.14 support include:
- Updating tests to reflect current Python behavior around decorated special methods
- Formally adding Python 3.14 to CI, packaging metadata, and documentation

Together, these changes allow FIREWHEEL to run and test cleanly under Python 3.14 while also improving maintainability of the affected code.

## Type of Change ##
- Other (please describe): Python 3.14 compatibility and internal refactoring

## Checklist ##

## Additional Notes ##
Notable implementation details:
- The `ExperimentGraph` worker refactor precomputes filtered vertex IDs in the parent process and sends only graph IDs to workers.
- The minimega compatibility check now uses a dedicated worker plus a safe process context manager to ensure proper termination and queue cleanup on timeout or error.
- The decorator-related unit test was updated to reflect Python’s special-method lookup semantics, which are especially relevant when validating behavior under newer Python versions.
- Project metadata and CI now explicitly include Python 3.14 support.

### DIFF
--- a/src/firewheel/lib/utilities.py
+++ b/src/firewheel/lib/utilities.py
@@ -6,10 +6,13 @@ import filecmp
 import hashlib
 import tarfile
 import traceback
+import multiprocessing
 from time import sleep
 from typing import Any, Tuple, Optional
 from pathlib import Path
 from functools import wraps as _wraps
+from contextlib import contextmanager
+from collections.abc import Generator
 
 from rich.console import Console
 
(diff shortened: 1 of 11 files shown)


src/firewheel/lib/utilities.py
```diff
@@ -451,3 +454,38 @@ def f_retry(*args: Any, **kwargs: Any):
...
+
+    Yields:
+        tuple[multiprocessing.Process, multiprocessing.Queue]: A pair
+            containing the spawned process instance and the queue
+            injected into the process.
+    """
+    context = multiprocessing.get_context()
+    task_queue = context.Queue()
+
+    # Launch a worker process to updae the queue
```
sdelliot:
```suggestion
    # Launch a worker process to update the queue
```

sdelliot (changes requested):
I think once we fix the minor spelling issue, this is good to go.

PR #289, opened 2026-06-12
test: Adding unittests for the `lib` module to increase code coverage and reliability

sdelliot:
## Description
This pull request adds a comprehensive unit test suite for the `firewheel.lib` package using `pytest`, following the general style of the existing FIREWHEEL unit tests.

The new tests cover core utility and infrastructure modules in `firewheel.lib`, including:
- `firewheel.lib.discovery.api`
- `firewheel.lib.experiment_utils`
- `firewheel.lib.grpc.firewheel_grpc_client`
- `firewheel.lib.grpc.firewheel_grpc_resources`
- `firewheel.lib.grpc.firewheel_grpc_server`
- `firewheel.lib.log`
- `firewheel.lib.minimega.api`
- `firewheel.lib.minimega.file_store`
- `firewheel.lib.utilities`

These tests exercise both standard behavior and error-handling branches, with a particular focus on:
- external dependency mocking (`requests`, gRPC, minimega, subprocess)
- malformed response and failure-path handling
- file/archive handling and cleanup logic
- lock, retry, and polling behavior
- regression coverage for discovered bugs in `FileStore`

In support of these tests, generated gRPC files were excluded from coverage evaluation so that coverage metrics reflect maintainable handwritten code rather than generated artifacts.

The `firewheel.lib` package previously had little to no dedicated unit test coverage despite containing a large amount of core functionality used by CLI, experiment save/load, minimega integration, discovery integration, logging, and gRPC coordination.

This change was required to improve confidence in the correctness of foundational library code, increase regression protection, and bring `lib` coverage in line with the rest of the codebase.

While developing these tests, the work also exposed real defect paths in `firewheel.lib.minimega.file_store`, including:
- incorrect use of `rstrip()` for suffix removal in `_strip_extension()`
- incomplete decompression exception handling for malformed `.xz` and tar content

The test suite was expanded to capture those cases so they remain protected going forward.

## Type of Change ##
- Other (please describe): test infrastructure / unit test coverage expansion

## Checklist ##


src/firewheel/tests/unit/lib/test_lib_discovery_api.py
```diff
@@ -0,0 +1,638 @@
...
+    response = Mock()
+    response.status_code = status_code
+    response.ok = ok
+    response.json.return_value = json_data
+    response.raise_for_status = Mock()
+    return response
+
+
+@patch("firewheel.lib.discovery.api.Config")
+def test_init_uses_config_defaults(mock_config_cls) -> None:
```
mitchnegus: Since this `mock_config_cls` seems to get used for the vast majority of tests without modification, my suggestion would be to use a fixture.

src/firewheel/tests/unit/lib/test_lib_grpc_resources.py
```diff
@@ -0,0 +1,19 @@
...
+
+from __future__ import annotations
+
+from unittest.mock import patch
+
+from firewheel.lib.grpc.firewheel_grpc_resources import msg_to_dict
+
+
+def test_msg_to_dict_replaces_none_string() -> None:
+    """Verify literal 'None' values are converted to Python None."""
```
mitchnegus:
```suggestion
    """Verify literal 'None' values are converted to Python :py:data:`None.`"""
```

(I was not going to change all the docstrings in all of these tests to necessarily use code styling, but in this case I thought it might help distinguish the literal 'None' from Python's `None` type.)

src/firewheel/tests/unit/lib/test_lib_minimega_api.py
```diff
@@ -0,0 +1,461 @@
...
+    api.log = Mock()
+    api.mm = Mock()
+    api.mm_base = "/tmp/mm"
+    api.mm_socket = "/tmp/mm/minimega"
+    return api
+
+
+def test_get_head_node(monkeypatch) -> None:
+    """Verify the head node is read from configuration."""
+    from firewheel.config import config
```
mitchnegus: Importing `config` on each test feels to me like a pattern we should avoid. If it's required for the monkeypatch to work, then maybe we also move this to a fixture so we do it once.

mitchnegus (commented):
This looks really good. My biggest suggestion would be to use pytest fixtures to reduce some of the unit test setup boilerplate (e.g., repeated calls to helper functions). 

sdelliot/firewheel#8 covers this adjustment in the majority of cases I saw, where setup was test independent and could be easily replaced via fixture.

src/firewheel/tests/unit/lib/test_lib_minimega_api.py
sdelliot: Totally agree! Your PR included this as a fixture and that seems like the right balance.

mitchnegus (approved):

PR #309, opened 2026-07-07
feat: Update key subset of helpers to provide JSON output

sdelliot:
## Description
Refactor several FIREWHEEL CLI helpers to support structured machine-readable output for future MCP integration while preserving existing human-facing behavior by default.

This PR introduces a consistent output pattern across the updated helpers:
- explicit `--format ...` selection for human-readable vs machine-readable output
- structured internal data collection separated from rendering
- Rich-based human rendering where appropriate
- JSON output for machine consumers
- backward-compatible legacy flags retained where needed (for example `--csv` and `--json` aliases)

The helpers updated in this PR are:
- `vm mix`
- `status`
- `repository list`
- `mm show_caches`
- `push file`
- `pull file`
- `time`
- `experiment`
- `vm list`
- `vm log`
- `mc list`

In addition to helper code changes, the corresponding helper documentation was updated to describe the new output modes and retained compatibility flags.

This change is required to make FIREWHEEL CLI helpers easier to integrate into a future MCP server and other automation tooling.

Previously, many helpers only emitted human-oriented text, tables, or ad hoc output that was difficult to consume programmatically. Even helpers with some structured output support did not consistently follow a shared pattern. This PR addresses that by standardizing helper behavior around:
- collecting structured result data first
- rendering either human-readable output or JSON from that same data
- preserving default UX for existing users

A few notable design points:
- `vm list` now supports `--format table|json|csv` while retaining `--csv` for backward compatibility.
- `vm log` now separates output format from artifact selection via `--format {text,json}` and `--log-type {text,json}`, while retaining `--json` as a backward-compatible alias for the JSON log artifact.
- `experiment` now supports `--format table|json`, returns structured JSON summaries, and uses safer per-phase output capture for JSON mode to reduce interference from helper/runtime output. JSON mode currently requires `--no-install` to avoid interactive Model Component INSTALL prompting.
- `mc list` was refactored to return structured grouped data and now uses Rich for text rendering rather than `colorama`, eliminating the need for the old monochrome flag.

This PR is intentionally scoped to a single goal: establishing consistent machine-readable output support and documentation updates across the targeted CLI helpers.

## Type of Change ##
- New feature
- Documentation update
- Other: CLI refactor for consistent machine-readable output conventions and backward-compatible output modernization

## Checklist ##

## Additional Notes ##
This PR completes an initial helper refactor set for MCP-oriented integration.

A possible follow-on PR on `experiment` may include:
- A better method of not requiring the `--no-install` flag.

Backward compatibility notes:
- `vm list --csv` is retained as an alias for CSV output
- `time --json` is retained as an alias for JSON output
- `vm log --json` is retained as an alias for selecting the JSON log artifact

Potential reviewer focus areas:
- structured output schemas and naming consistency
- `experiment` JSON-mode capture behavior
- `vm log` multiline text log parsing and JSON log artifact handling
- documentation alignment with final helper behavior

### DIFF
--- a/src/firewheel/cli/helpers/mm/show_caches
+++ b/src/firewheel/cli/helpers/mm/show_caches
@@ -2,15 +2,37 @@ AUTHOR
 FIREWHEEL Team
 DONE
 DESCRIPTION
-
-List the image and vm_resource cache directories which are found in the
+List the image and vm_resource cache directories found in the
 :class:`FileStore <firewheel.lib.minimega.file_store.FileStore>`. This enables
-users to identify if a file was correctly cached.
+users to identify whether a file was correctly cached.
+
+This Helper supports both human-readable table output and machine-readable
+JSON output.
+
+Arguments
++++++++++
+
+Named Arguments
+^^^^^^^^^^^^^^^
+
+.. option:: -h, --help
+
+    Show help message and exit.
+
+.. option:: --format {text,json}
+
+    :default: ``text``
+
+    Select the output format. ``text`` displays human-readable cache tables with
+    file sizes rendered in a human-friendly format. ``json`` returns a structured
+    JSON response with raw byte counts.
 
 Examples
 ++++++++
 
 ``firewheel mm show_caches``
+
+``firewheel mm show_caches --format json``
 
 DONE
 RUN LocalPython ON control
--- a/src/firewheel/cli/helpers/mm/show_caches
+++ b/src/firewheel/cli/helpers/mm/show_caches
@@ -106,8 +106,8 @@ def collect_cache_data():
 
     host = vm_resources_host or images_host or mm_api.get_head_node()
 
-    vm_resources_entries = normalize_cache_rows(vm_resources_rows)
-    images_entries = normalize_cache_rows(images_rows)
+    vm_resources_entries = sorted(normalize_cache_rows(vm_resources_rows), key=lambda r: r["name"])
+    images_entries = sorted(normalize_cache_rows(images_rows), key=lambda r: r["name"])
 
     return {
         "host": host,
--- a/src/firewheel/cli/helpers/mm/show_caches
+++ b/src/firewheel/cli/helpers/mm/show_caches
@@ -13,17 +13,206 @@ Examples
 ``firewheel mm show_caches``
 
 DONE
-RUN Shell ON control
-#!/bin/bash
-MM_INSTALL_DIR="$("$FIREWHEEL_PYTHON" "$FIREWHEEL" config get minimega.install_dir)"
-MM_BASE_DIR="$("$FIREWHEEL_PYTHON" "$FIREWHEEL" config get minimega.base_dir)"
-MINIMEGA_BIN="$MM_INSTALL_DIR/bin/minimega -base=$MM_BASE_DIR"
+RUN LocalPython ON control
+#!/usr/bin/env python
 
-HOST=$(hostname)
-echo "Cache contents from $HOST:vm_resources"
-$MINIMEGA_BIN -e file list vm_resources
-echo
-echo "Cache contents from $HOST:images"
-$MINIMEGA_BIN -e file list images
-echo
+import json
+import argparse
+from datetime import datetime, timezone
+
+from rich.console import Console
+
+from firewheel.cli.utils import RichDefaultTable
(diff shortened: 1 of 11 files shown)


sdelliot:
We should attempt to add FIREWHEEL restart to this set if feasible.

sdelliot:
> We should attempt to add FIREWHEEL restart to this set if feasible.

There isn't currently an easy way to do this. We should proceed with this PR and address that separately.

src/firewheel/cli/helpers/mm/show_caches
```diff
@@ -2,28 +2,239 @@ AUTHOR
...
+    table = RichDefaultTable(
+        title=f"{title} ({host})",
+        caption=f"Found [cyan]{len(rows)}[/] entries",
+    )
+    table.add_column("Name")
+    table.add_column("Size", justify="right")
+    table.add_column("Modified")
+
+    if not rows:
+        table.add_row("N/A", "N/A", "0 B", "N/A")
```
mitchnegus: Seem to be two four values here. Is the `"0 B"` unnecessary?

src/firewheel/cli/helpers/vm/log
```diff
@@ -31,83 +54,379 @@ Positional Arguments
...
+                    "continuation": [],
+                }
+            else:
+                current_entry["continuation"].append(line)
+
+    if current_entry is not None:
+        current_entry["raw_message"] = "\n".join(
+            [current_entry["message"], *current_entry["continuation"]]
+        )
+        entries.append(current_entry)
```
mitchnegus: So it looks like `continuation` gets passed out of this function even though it's just an intermediate artifact. That's easy enough to fix with a call to `.pop("continuation")`. It took me a bit to grok this whole section though, and I'm wondering if we might split it into a chunk that just collects entries, and then another returning the formatted version.


```suggestion
    RawLogEntry = namedtuple("RawLogEntry", ["timestamp", "level", "lines"]
    raw_entries = []

    for line in text.splitlines():
        if is_new_log_line := LOG_LINE_RE.match(line):
            raw_entry = RawLogEntry(
                timestamp=match.group("timestamp"), 
                level=match.group("level"),
                lines=[match.group("message")],
            )
            raw_entries.append(raw_entry)
        else:
            if raw_entries:
                raw_entries[-1].lines.append(line)
            else:
                # Handle edge cases where no prior entry exists, but the line is not prefixed
                raw_entries.append(RawLogEntry(timestamp=None, level=None, lines=[line]))
    return [
        {
            "timestamp": entry.timestamp,
            "level": entry.level,
            "message": entry.lines[0],
            "raw_message": "\n".join(entry.lines),
         }
         for entry in raw_entries
    ]
```

This format also makes clear that the return value has the message set to just the first line of the logged output, and the raw message is the full thing. That appears to be the desired behavior, where we often have the message be the string-output of the error, and the remainder be the traceback.

mitchnegus: I also haven't tested or checked formatting of that block since it's here in the browser suggestions, so if this is a change worth making I can do some better checking to save debugging.

src/firewheel/cli/helpers/vm/list
```diff
@@ -120,195 +128,426 @@ def vm_list(args):
...
+    label_map = {
+        "name": "Name",
+        "uuid": "UUID",
+        "state": "State",
+        "image": "Image",
+        "hostname": "Hostname",
+        "vnc": "VNC Port",
+        "time": "Time",
+        "ip": "Control IP",
+    }
```
mitchnegus: Since this `label_map` and the `header_map` in `render_csv` (below) are identical, and probably always should be identical, maybe we make them a module level constant like `FIELD_LABELS` or `FIELD_TITLES`?

mitchnegus (commented):
This looks good and works for me when I test it. I agree, the `--no-install` option is not intuitive (I almost raised that as an issue before I noticed that you called it out). Maybe we add like a `--non-interactive` option that just runs the experiment (but fails out if a MC is not installed). Either way, I agree, let's defer that to another issue.

One point maybe worth discussing, since you had asked about the structure of output schemas. It looks like some of the outputs have `"success"` and `"error"` keys that are not universal. Should we make those universal? 

There are likely some other things that I may follow-up on in subsequent MRs, but nothing that is worth holding this up further.

PR #322, opened 2026-08-13
chore: Ignore various AI-related files

sdelliot:
## Description
With so many AI tools starting to code, we should proactively ignore these AI files that could pollute our repo.

## Type of Change ##

- Other (please describe): Chore

## Checklist ##

~~- [ ] I have made corresponding changes to the documentation.~~
~~- [ ] I have tested my code with the functional test suite, and it passed.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -221,6 +221,50 @@ tags
 
 # End of https://www.toptal.com/developers/gitignore/api/python,vim,macos
 
+# ==========================================
+# AI Coding Assistants & Editors
+# ==========================================
+
+# Codex & OpenCode
+.codex/
+.opencode/
+codex_config.*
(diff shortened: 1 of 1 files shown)


GhostofGoes:
Could you apply this to minimega as well?

GhostofGoes (approved):

mitchnegus:
I will note that some of the directories here may exclude files that we could at some point want—for example, I believe that some teams may intentionally commit files like `.claude/settings.json`. 

Then again, I am 100% in favor of approving this now and adding those back in when we want them. We can then be sure that they are things we actually want to commit on a case-by-case basis. I see zero rush.

mitchnegus (approved):

